PR #43, opened 2015-02-18
Fix decompressFromUint8Array, replace .apply() with forEach

carlansley:
The original .apply() was overflowing the call stack.  Switched to a forEach.


pieroxy:
Did you run the unit tests on this one ?

carlansley:
Yeah works in Safari, Chrome, Firefox & Opera.  Didn't get a chance to test IE but if for versions that support typed arrays it should be fine.

Issue #47, opened 2015-03-14
Use ES6 Map/Set when possible.


ChALkeR:
Using ES6 `Map` for `context_dictionary` and ES6 `Set` (or `Map`) for `context_dictionaryToCreate` instead of a plain Object speeds up the compression twice in v8.

See https://github.com/pieroxy/lz-string/issues/46#issuecomment-80531018 for test results.

`Map` and `Set` are supported in current Chrome and Firefox versions.
IE11 has partial support for them, but that should be enough: see «Map → basic functionality» here: https://kangax.github.io/compat-table/es6/

I haven't tested the speed in Firefox and IE11, but it shouldn't be worse than using an Object.

Supporting old browsers would require some kind of feature detection.

pieroxy:
Thanks for the idea and the tests. I've always wanted lz-stringto be compliant with every javascript engine, hence the avoidance of new JavaScript stuff. I've also done a  [few tests](http://pieroxy.net/blog/pages/lz-string/testing.html) to work with arrays instead of Strings but I probably didn't push them far enough.

If you have the code with ES6 Map/Set, can you push it through a pull request ? I can probably have a look into it this week. Detecting the feature shouldn't be that hard.

ALSO: Can you take a minute and pass the jsperf of the latest pre-release (https://github.com/pieroxy/lz-string/releases) with all the IEs you have at hand? I can't decide myself to release it (faster on FF/Chrome, slower on IE) but I lack data on IE and my tests were performed in a VM which is probable messing with performances.

ChALkeR:
I don't have any IE versions, sorry.
What about #46? Should I make a pull request for that? That one is more severe than this.

pieroxy:
I've had my share of surprises on jsperf with code that would OBVIOUSLY be faster but was slower on some browsers for reasons I couldn't be bothered to investigate.

I'd start with doing a couple of jsperf with this one as it looks like it's going to be beneficial for #46 as well. We'll move forward if tests are a clear win on all platform supported. Moving forward meaning:
 * Feature detection
 * implement the new version if feature is there while keeping the actual for old browsers
 * We need to be careful not to break anything in node.js as well.

NOTE: I've just released v1.4.0 which clean up the wrapper methods. If you can rebase on top of that it would be great.

Either you do the jsperf yourself or you push a pull request and I'll deal with the jsperf.

ChALkeR:
This one won't help #46.
The problem with #46 is the memory usage, jsperf doesn't measure that.

[...]

PR #49, opened 2015-03-23
Do not export private objects.

ChALkeR:
A little bit cleaner code (inside functions).
Should do a bit less object property accessing.

The main point of this patch is that it adds a place where the Map/Set fallback for #47 could be placed without polluting global scope in browsers: in an if-then-else just above the LZString object.

Should be of the same speed or slightly faster in all browsers. If it is not, something is wrong.

### DIFF
--- a/libs/lz-string.js
+++ b/libs/lz-string.js
@@ -7,26 +7,28 @@
 // http://pieroxy.net/blog/pages/lz-string/testing.html
 //
 // LZ-based compression algorithm, version 1.4.1
-var LZString = {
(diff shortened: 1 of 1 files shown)


pieroxy:
Hi

Thanks for the pull request. I'll just make a small jsperf to make sure everything is ok and I'll merge it right away. I did not notice vars were global... I guess it was too obvious for me to see.

I just released an ugly patch which apparently conflicts with yours. I'll take care of all that this weekend.

PR #51, opened 2015-04-14
Added TypeScript definition for lz-string module

axel-wilczek:
Added a TypeScript definition file (*.d.ts) for the lz-string main library. The interface is pretty basic but provides IntelliSense support and type safety.

https://typescript.codeplex.com/wikipage?title=Writing%20Definition%20%28.d.ts%29%20Files

I am thinking about adding typings to the lz-string library itself. What are your feelings about this step? Before I put any work into this, I want to be sure it is welcome.


pieroxy:
Thanks for the pull request. I'll do a release most likely this weekend.

PR #52, opened 2015-04-15
Add support for in-browser use with AMD loaders

zanona:
This change allows using the library from within the browser when using AMD loaders such as require.js while still maintaining commonJS calls intact.


pieroxy:
I will do a release most likely this weekend

zanona:
Thank you very much

PR #66, opened 2015-11-04
Update lz-string.js

darrylyeo:
Version 1.5.0, perhaps?
* Made code _much_ more compact and reusable (reduced **501** lines of code to **273**)!
* Defined some reusable functions within `_compress()` (perhaps you can come up with better names).

### DIFF
--- a/libs/lz-string.js
+++ b/libs/lz-string.js
@@ -8,473 +8,248 @@
 //
 // LZ-based compression algorithm, version 1.4.4
 var LZString = (function() {
-
 // private property
-var f = String.fromCharCode;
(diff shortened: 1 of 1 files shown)


pieroxy:
If you look at LZ-String history, in inlined all the functions on purpose to make it faster... I know it made the code awful and bigger, but the performance was greatly increased.

Ca you do a jsperf to compare it to the current version ?

darrylyeo:
Okay, I moved the three helper functions to outside `_compress()` so as to not redefine them each time. It comes out to be a little slower.
http://jsperf.com/lzstring-test

pieroxy:
Well, it doesn't look good. Am I willing to trade a few lines vs 20% slower compression? Well, not really... 

On another note, I couldn't test on internet explorer for some reason...

darrylyeo:
I see what you mean.

PR #81, opened 2016-11-13
TypeScript: adjust reference module declaration

jhuss:
importing lz-script on TypeScript:

`/// <reference path="node_modules/lz-string/typings/lz-string.d.ts" />`
`import * as LZString from 'LZString';`

 is getting error:

> error TS2307: Cannot find module 'LZString'

now, checking documentation (https://www.typescriptlang.org/docs/handbook/namespaces-and-modules.html#pitfalls-of-namespaces-and-modules) and updating the format in lz-string.d.ts, the TS compiler it's not showing anymore the import error

### DIFF
--- a/typings/lz-string.d.ts
+++ b/typings/lz-string.d.ts
@@ -1,16 +1,16 @@
-declare module LZString {
-  function compressToBase64(input: string): string;
-  function decompressFromBase64(input: string): string;
+declare module "LZString" {
+  export function compressToBase64(input: string): string;
+  export function decompressFromBase64(input: string): string;
 
-  function compressToUTF16(input: string): string;
-  function decompressFromUTF16(compressed: string): string;
+  export function compressToUTF16(input: string): string;
+  export function decompressFromUTF16(compressed: string): string;
 
-  function compressToUint8Array(uncompressed: string): Uint8Array;
-  function decompressFromUint8Array(compressed: Uint8Array): string;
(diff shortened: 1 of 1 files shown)


jhuss:
maybe the <reference> comment is missing?, this was working with typings tool.
now I resolved with npm install @types/lz-string (for newer TypeScript versions)
your PR looks and works fine to me, I will close this PR

jhuss:
PR #83 works well, and this no longer needed in favour of @types/lz-string

janaagaard75:
There is no need to install @types/lz-string, since this repo now already has the require type definitions. @types/lz-string should be marked as deprecated.

jhuss:
right, I referred that no longer needed this PR, not the type definition.
while lz-string does not include the corrections, a valid option is to use @types/lz-string.

pieroxy:
So, I didn't follow this closely enough, sorry about that. Is there anything else needed in lz-string ?

janaagaard75:
I think the type definition from Definitely Typed (@typez/lz-string) should be copied to this repo, and then actually removed from Definitely Typed, since they are no longer there there, but I don't now if this is just something that you do, or if you ask kindly for permissions first. The readme in @typez/lz-string should then be updated with a note, explaining that the type definitions now included lz-string, so it's no longer necessary to add @types/lz-string separately.

PR #83, opened 2017-02-15
Make typings work out-of-the-box

janaagaard75:
This makes the typings work out-of-the-box.

```typescript
import * as LZString from 'lz-string';
const compressed = LZString.compressToBase64('My string');
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "string"
   ],
   "main": "libs/lz-string.js",
+  "typings": "typings/lz-string.d.ts",
   "bin": {
     "lz-string": "bin/bin.js"
   },
--- a/typings/lz-string.d.ts
+++ b/typings/lz-string.d.ts
@@ -1,16 +1,14 @@
-declare module LZString {
-  function compressToBase64(input: string): string;
-  function decompressFromBase64(input: string): string;
+export function compressToBase64(input: string): string;
+export function decompressFromBase64(input: string): string;
 
-  function compressToUTF16(input: string): string;
-  function decompressFromUTF16(compressed: string): string;
+export function compressToUTF16(input: string): string;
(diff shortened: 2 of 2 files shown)


janaagaard75:
I don't know why there isn't any updates on this. If you want to use lz-string in your TypeScript project today, you can replace `"lz-string": "1.4.4"` with the following in your `package.json`:

    "lz-string": "git://github.com/janaagaard75/lz-string"

Drezounet:
> I don't know why there isn't any updates on this. If you want to use lz-string in your TypeScript project today, you can replace "lz-string": "1.4.4" with the following in your package.json:

"lz-string": "git://github.com/janaagaard75/lz-string"

this doesn't work :
Unsupported command arguments: -c core.longpaths=true clone --temp [...] etc...

janaagaard75:
What command are you using, and what version of npm are you running, @Drezounet? `npm install` runs fine on this `package.json` using npm version 4.1.2.

https://github.com/janaagaard75/film-filter/blob/master/website/package.json

Drezounet:
finally it's ok. I clean up everything and now work's like a charm !
i've some error in VS : module lz-string not found. 
But, build ok and run fine.

janaagaard75:
Thanks for for correctly the style of the type definitions. This was one of my first attempts at this. :)

I believe there is now a `@types/lz-string` package that could be used. Personally I prefer when the type definitions are part of the package, though, so somebody might have to create a `lz-string-ts` (or `ts-lz-string`) package. :-/

thekip:
Yeah, you are right, and [@types/lz-string](https://www.npmjs.com/package/@types/lz-string) this typing even more "correct" than mine. 

I actually was confused when found typings folder in project, and after that even haven't tried to look at `@types`

Issue #88, opened 2017-03-24
micro-optimisation: use array lookup instead of charAt?


JobLeonard:
Hey, I was just looking through the `base64-string` code out of curiosity for how you did it, when I spotted that you use `charAt` and `indexOf` on a string:
https://github.com/pieroxy/lz-string/blob/master/libs/base64-string.js#L178

My intuition tells me that this is slower than using an array of single character string, so I made these little tests to see if that is true:

http://jsbench.github.io/#cbfad9d07c549532aceb79f8b051553e

On my laptop, encoding via array lookup is _slightly_ faster than using `charAt` on a string. Decoding is slower (I suspect that for larger strings the binary search hybrid migh have worked, but 64 characters is just so small a list that linear search + under-the-hood optimisations are faster).

There's also the possibility that this slightly reduces memory usage - I don't know if charAt creates a new string every time or not, but using the array method would imply pushing the same immutable string every time, right? (although even if there is a difference, it is probably insignificant).

JobLeonard:
I just updated the tests with a dictionary decode as well, [since that is what you use elsewhere too](https://github.com/pieroxy/lz-string/blob/master/libs/lz-string.js#[L14-L26]). It's almost 2.5x as fast on my desktop browsers, and in mobile Chrome but no Firefox (identified as "Android 52 browser" in the charts for some reason).

Again, I know it's a micro-op in a larger context so I totally understand if you wouldn't consider this worth changing the code for, but it's nice to have some up-to-date tests at least, no? (assuming I made no mistakes in the code and the benchmarks are valid)

pieroxy:
It's interesting. Could you push a MR ?

tophf:
@JobLeonard, it'd be interesting to combine your approach with Map/Set as discussed in #47 (I'm using a locally modified version that's ~2 times faster overall).

JobLeonard:
@pieroxy Sure, I'll make a branch and try.

(Aside, anyone else annoyed by how both jsperf and jsbench repaint the DOM while doing pure-JS benchmarks? I had it lag on a few tests because of DOM reflow calculations, that surely skewed the numbers)

JobLeonard:
Ok, before I do this I have some more testing to do. 

You see, it occurred to me that we can also try pairs of characters. In that case, both the dictionary and the binary search options should end up significantly faster, and perhaps faster than the single-char options too.

[...]

PR #96, opened 2017-06-25
Base64 array/dict lookup optimisations

JobLeonard:
See issue #88. I didn't actually test if it was faster or not.


JobLeonard:
Wait, there's a few bugs. I'll try again later.

Issue #97, opened 2017-06-26
Avoid .join('') followed by .charAt() or .charCodeAt()


JobLeonard:
So [as proposed in this comment](https://github.com/pieroxy/lz-string/pull/80#issuecomment-310926428), with a fairly trivial refactoring we can change a lot of internal cases where we:

- construct an array of chars
- `.join('')` the array
- proceed to use `.charAt()` or `.charCodeAt()` on the resulting string

.. to leave out the middle step entirely, and replace `.charAt()` calls with array indexing. For cases where we use `.charCodeAt()`, what we have to do is replace `resultString.charCodeAt(i)` with `resultArray[i].charCodeAt()` - not entirely sure which is faster, but either way it saves the creation of a string. That in itself might save us enough performance to make it worth it, and if we're compressing lots of large data that could matter in memory overhead too.

There might also be a few use-cases in the wild where it's more practical for people to have a returned array of chars instead of a single string, so it could also be a useful (fully backwards compatible) addition to the API.

Like I said, I'm on sick leave at home and currently bored, so I'm going to work on this, both in the `lz-string.js` and `base64-string.js` implementations, and push a PR later. Then I leave it up to you to decide if you like the changes!

PR #98, opened 2017-06-26
Array lookup + avoid intermediate string creation

JobLeonard:
This combines various suggestions from @Rycochet and @lishid and myself. See #88 and #97.

- replace string-based `keyStr.charAt(i)` with array-based `keyStr[i]`
- add `compressToArray` and `decompressFromArray` functions, and use them throughout the code to avoid intermediate string creation:
  - `compressToBase64` avoids appending to string
  - `compressToUint8Array` avoids an intermediate string
  - `decompressFromUint8Array` avoids an intermediate string

Similar optimisations have been applied to `base64-string.js`, but it doesn't have tests so I can't say for sure if it's bug-free.

### DIFF
--- a/libs/base64-string.js
+++ b/libs/base64-string.js
@@ -9,6 +9,19 @@
 //
 // Base64 compression / decompression for already compressed content (gif, png, jpg, mp3, ...)
 // version 1.4.1
+
+var Base64String = (function(){
+var f = String.fromCharCode;
+var keyStrEnc = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=".split('');
+var keyStrDec = (function (){
+    var dict = {}
+    var i = keyStrEnc.length;
+    while(i--){
+      dict[keyStrEnc[i]] = i;
+    }
+    return dict;
+  })();
+
 var Base64String = {
 
   compressToUTF16 : function (input) {
@@ -23,68 +36,68 @@ var Base64String = {
       c = input.charCodeAt(i);
       switch (status++) {
         case 0:
-          output.push(String.fromCharCode((c >> 1)+32));
+          output.push(f((c >> 1)+32));
           current = (c & 1) << 14;
           break;
         case 1:
-          output.push(String.fromCharCode((current + (c >> 2))+32));
+          output.push(f((current + (c >> 2))+32));
           current = (c & 3) << 13;
           break;
         case 2:
-          output.push(String.fromCharCode((current + (c >> 3))+32));
+          output.push(f((current + (c >> 3))+32));
           current = (c & 7) << 12;
           break;
         case 3:
-          output.push(String.fromCharCode((current + (c >> 4))+32));
+          output.push(f((current + (c >> 4))+32));
           current = (c & 15) << 11;
           break;
         case 4:
-          output.push(String.fromCharCode((current + (c >> 5))+32));
+          output.push(f((current + (c >> 5))+32));
           current = (c & 31) << 10;
           break;
         case 5:
-          output.push(String.fromCharCode((current + (c >> 6))+32));
+          output.push(f((current + (c >> 6))+32));
           current = (c & 63) << 9;
           break;
         case 6:
-          output.push(String.fromCharCode((current + (c >> 7))+32));
+          output.push(f((current + (c >> 7))+32));
           current = (c & 127) << 8;
           break;
         case 7:
-          output.push(String.fromCharCode((current + (c >> 8))+32));
+          output.push(f((current + (c >> 8))+32));
           current = (c & 255) << 7;
           break;
         case 8:
-          output.push(String.fromCharCode((current + (c >> 9))+32));
+          output.push(f((current + (c >> 9))+32));
           current = (c & 511) << 6;
           break;
         case 9:
-          output.push(String.fromCharCode((current + (c >> 10))+32));
+          output.push(f((current + (c >> 10))+32));
           current = (c & 1023) << 5;
           break;
         case 10:
-          output.push(String.fromCharCode((current + (c >> 11))+32));
+          output.push(f((current + (c >> 11))+32));
           current = (c & 2047) << 4;
           break;
         case 11:
-          output.push(String.fromCharCode((current + (c >> 12))+32));
+          output.push(f((current + (c >> 12))+32));
           current = (c & 4095) << 3;
           break;
         case 12:
-          output.push(String.fromCharCode((current + (c >> 13))+32));
+          output.push(f((current + (c >> 13))+32));
           current = (c & 8191) << 2;
           break;
         case 13:
-          output.push(String.fromCharCode((current + (c >> 14))+32));
+          output.push(f((current + (c >> 14))+32));
           current = (c & 16383) << 1;
           break;
         case 14:
-          output.push(String.fromCharCode((current + (c >> 15))+32, (c & 32767)+32));
+          output.push(f((current + (c >> 15))+32, (c & 32767)+32));
           status = 0;
           break;
       }
     }
-    output.push(String.fromCharCode(current + 32));
+    output.push(f(current + 32));
     return output.join('');
   },
 
@@ -103,63 +116,63 @@ var Base64String = {
           current = c << 1;
           break;
         case 1:
-          output.push(String.fromCharCode(current | (c >> 14)));
+          output.push(f(current | (c >> 14)));
           current = (c&16383) << 2;
           break;
         case 2:
-          output.push(String.fromCharCode(current | (c >> 13)));
+          output.push(f(current | (c >> 13)));
           current = (c&8191) << 3;
           break;
         case 3:
-          output.push(String.fromCharCode(current | (c >> 12)));
+          output.push(f(current | (c >> 12)));
           current = (c&4095) << 4;
           break;
         case 4:
-          output.push(String.fromCharCode(current | (c >> 11)));
+          output.push(f(current | (c >> 11)));
           current = (c&2047) << 5;
           break;
         case 5:
-          output.push(String.fromCharCode(current | (c >> 10)));
+          output.push(f(current | (c >> 10)));
           current = (c&1023) << 6;
           break;
         case 6:
-          output.push(String.fromCharCode(current | (c >> 9)));
+          output.push(f(current | (c >> 9)));
           current = (c&511) << 7;
           break;
         case 7:
-          output.push(String.fromCharCode(current | (c >> 8)));
+          output.push(f(current | (c >> 8)));
           current = (c&255) << 8;
           break;
         case 8:
-          output.push(String.fromCharCode(current | (c >> 7)));
+          output.push(f(current | (c >> 7)));
           current = (c&127) << 9;
           break;
         case 9:
-          output.push(String.fromCharCode(current | (c >> 6)));
+          output.push(f(current | (c >> 6)));
           current = (c&63) << 10;
           break;
         case 10:
-          output.push(String.fromCharCode(current | (c >> 5)));
+          output.push(f(current | (c >> 5)));
           current = (c&31) << 11;
           break;
         case 11:
-          output.push(String.fromCharCode(current | (c >> 4)));
+          output.push(f(current | (c >> 4)));
           current = (c&15) << 12;
           break;
         case 12:
-          output.push(String.fromCharCode(current | (c >> 3)));
+          output.push(f(current | (c >> 3)));
           current = (c&7) << 13;
           break;
         case 13:
-          output.push(String.fromCharCode(current | (c >> 2)));
+          output.push(f(current | (c >> 2)));
           current = (c&3) << 14;
           break;
         case 14:
-          output.push(String.fromCharCode(current | (c >> 1)));
+          output.push(f(current | (c >> 1)));
           current = (c&1) << 15;
           break;
         case 15:
-          output.push(String.fromCharCode(current | c));
+          output.push(f(current | c));
           status=0;
           break;
       }
@@ -172,20 +185,6 @@ var Base64String = {
     //return output;
 
   },
-
-
-  // private property
-  _keyStr : "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=",
-  _keyStrDec : this._keyStr.split(''),
-  _keyStrEnc : (function (){
-    // IIFE to create a dictionary
-    var dict = {}
-    var i = this._keyStrDec.length;
-    while(i--){
-      dict[this._keyStrDec[i]] = i;
-    }
-    return dict;
-  })(),
 
   decompress : function (input) {
     var output = [];
@@ -223,10 +222,10 @@ var Base64String = {
         enc4 = 64;
       }
 
-      output.push(this._keyStrDec[enc1]);
-      output.push(this._keyStrDec[enc2]);
-      output.push(this._keyStrDec[enc3]);
-      output.push(this._keyStrDec[enc4]);
+      output.push(keyStrDec[enc1]);
+      output.push(keyStrDec[enc2]);
+      output.push(keyStrDec[enc3]);
+      output.push(keyStrDec[enc4]);
     }
 
     return output.join('');
@@ -244,10 +243,10 @@ var Base64String = {
 
     while (i < input.length) {
 
-      enc1 = this._keyStrEnc[input.charAt(i++)];
-      enc2 = this._keyStrEnc[input.charAt(i++)];
-      enc3 = this._keyStrEnc[input.charAt(i++)];
-      enc4 = this._keyStrEnc[input.charAt(i++)];
+      enc1 = keyStrEnc[input.charAt(i++)];
+      enc2 = keyStrEnc[input.charAt(i++)];
+      enc3 = keyStrEnc[input.charAt(i++)];
+      enc4 = keyStrEnc[input.charAt(i++)];
 
       chr1 = (enc1 << 2) | (enc2 >> 4);
       chr2 = ((enc2 & 15) << 4) | (enc3 >> 2);
@@ -258,7 +257,7 @@ var Base64String = {
         flush = true;
 
         if (enc3 != 64) {
-          output.push(String.fromCharCode(output_ | chr2));
+          output.push(f(output_ | chr2));
           flush = false;
         }
         if (enc4 != 64) {
@@ -266,7 +265,7 @@ var Base64String = {
           flush = true;
         }
       } else {
-        output.push(String.fromCharCode(output_ | chr1));
+        output.push(f(output_ | chr1));
         flush = false;
 
         if (enc3 != 64) {
@@ -274,7 +273,7 @@ var Base64String = {
           flush = true;
         }
         if (enc4 != 64) {
-          output.push(String.fromCharCode(output_ | chr3));
+          output.push(f(output_ | chr3));
           flush = false;
         }
       }
@@ -282,14 +281,24 @@ var Base64String = {
     }
 
     if (flush) {
-      output.push(String.fromCharCode(output_));
+      output.push(f(output_));
       output = output.join('');
-      output = String.fromCharCode(output.charCodeAt(0)|256) + output.substring(1);
+      output = f(output.charCodeAt(0)|256) + output.substring(1);
     } else {
       output = output.join('');
     }
 
     return output;
 
-  }
+  },
+
+  compressToArray : function (input) {
+
+  },
+
+  decompressToArray : function (input) {
+
+  },
 }
+return Base64String;
+})()
--- a/libs/base64-string.js
+++ b/libs/base64-string.js
@@ -7,18 +7,18 @@
 // For more information, the home page:
 // http://pieroxy.net/blog/pages/lz-string/index.html
 //
-// Base64 compression / decompression for already compressed content (gif, png, jpg, mp3, ...) 
+// Base64 compression / decompression for already compressed content (gif, png, jpg, mp3, ...)
 // version 1.4.1
 var Base64String = {
-  
+
   compressToUTF16 : function (input) {
     var output = [],
         i,c,
         current,
         status = 0;
-    
+
     input = this.compress(input);
-    
+
     for (i=0 ; i<input.length ; i++) {
       c = input.charCodeAt(i);
       switch (status++) {
@@ -87,17 +87,17 @@ var Base64String = {
     output.push(String.fromCharCode(current + 32));
     return output.join('');
   },
-  
+
 
   decompressFromUTF16 : function (input) {
     var output = [],
         current,c,
         status=0,
         i = 0;
-    
+
     while (i < input.length) {
       c = input.charCodeAt(i) - 32;
-      
+
       switch (status++) {
         case 0:
           current = c << 1;
@@ -163,90 +163,100 @@ var Base64String = {
           status=0;
           break;
       }
-      
-      
+
+
       i++;
     }
-    
+
     return this.decompress(output.join(''));
     //return output;
-    
+
   },
 
 
   // private property
   _keyStr : "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=",
-  
+  _keyStrDec : this._keyStr.split(''),
+  _keyStrEnc : (function (){
+    // IIFE to create a dictionary
+    var dict = {}
+    var i = this._keyStrDec.length;
+    while(i--){
+      dict[this._keyStrDec[i]] = i;
+    }
+    return dict;
+  })(),
+
   decompress : function (input) {
     var output = [];
     var chr1, chr2, chr3, enc1, enc2, enc3, enc4;
     var i = 1;
     var odd = input.charCodeAt(0) >> 8;
-    
+
     while (i < input.length*2 && (i < input.length*2-1 || odd==0)) {
-      
+
       if (i%2==0) {
         chr1 = input.charCodeAt(i/2) >> 8;
         chr2 = input.charCodeAt(i/2) & 255;
-        if (i/2+1 < input.length) 
+        if (i/2+1 < input.length)
           chr3 = input.charCodeAt(i/2+1) >> 8;
-        else 
+        else
           chr3 = NaN;
       } else {
         chr1 = input.charCodeAt((i-1)/2) & 255;
         if ((i+1)/2 < input.length) {
           chr2 = input.charCodeAt((i+1)/2) >> 8;
           chr3 = input.charCodeAt((i+1)/2) & 255;
-        } else 
+        } else
           chr2=chr3=NaN;
       }
       i+=3;
-      
+
       enc1 = chr1 >> 2;
       enc2 = ((chr1 & 3) << 4) | (chr2 >> 4);
       enc3 = ((chr2 & 15) << 2) | (chr3 >> 6);
       enc4 = chr3 & 63;
-      
+
       if (isNaN(chr2) || (i==input.length*2+1 && odd)) {
         enc3 = enc4 = 64;
       } else if (isNaN(chr3) || (i==input.length*2 && odd)) {
         enc4 = 64;
       }
-      
-      output.push(this._keyStr.charAt(enc1));
-      output.push(this._keyStr.charAt(enc2));
-      output.push(this._keyStr.charAt(enc3));
-      output.push(this._keyStr.charAt(enc4));
+
+      output.push(this._keyStrDec[enc1]);
+      output.push(this._keyStrDec[enc2]);
+      output.push(this._keyStrDec[enc3]);
+      output.push(this._keyStrDec[enc4]);
     }
-    
+
     return output.join('');
   },
-  
+
   compress : function (input) {
     var output = [],
-        ol = 1, 
+        ol = 1,
         output_,
         chr1, chr2, chr3,
         enc1, enc2, enc3, enc4,
         i = 0, flush=false;
-    
+
     input = input.replace(/[^A-Za-z0-9\+\/\=]/g, "");
-    
+
     while (i < input.length) {
-      
-      enc1 = this._keyStr.indexOf(input.charAt(i++));
-      enc2 = this._keyStr.indexOf(input.charAt(i++));
-      enc3 = this._keyStr.indexOf(input.charAt(i++));
-      enc4 = this._keyStr.indexOf(input.charAt(i++));
-      
+
+      enc1 = this._keyStrEnc[input.charAt(i++)];
+      enc2 = this._keyStrEnc[input.charAt(i++)];
+      enc3 = this._keyStrEnc[input.charAt(i++)];
+      enc4 = this._keyStrEnc[input.charAt(i++)];
+
       chr1 = (enc1 << 2) | (enc2 >> 4);
       chr2 = ((enc2 & 15) << 4) | (enc3 >> 2);
       chr3 = ((enc3 & 3) << 6) | enc4;
-      
+
       if (ol%2==0) {
         output_ = chr1 << 8;
         flush = true;
-        
+
         if (enc3 != 64) {
           output.push(String.fromCharCode(output_ | chr2));
           flush = false;
@@ -258,7 +268,7 @@ var Base64String = {
       } else {
         output.push(String.fromCharCode(output_ | chr1));
         flush = false;
-        
+
         if (enc3 != 64) {
           output_ = chr2 << 8;
           flush = true;
@@ -270,7 +280,7 @@ var Base64String = {
       }
       ol+=3;
     }
-    
+
     if (flush) {
       output.push(String.fromCharCode(output_));
       output = output.join('');
@@ -278,8 +288,8 @@ var Base64String = {
     } else {
       output = output.join('');
     }
-    
+
     return output;
-    
+
   }
 }
--- a/libs/base64-string.js
+++ b/libs/base64-string.js
@@ -12,12 +12,12 @@
 
 var Base64String = (function(){
 var f = String.fromCharCode;
-var keyStrEnc = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=".split('');
-var keyStrDec = (function (){
-    var dict = {}
-    var i = keyStrEnc.length;
+var keyEnc = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=".split('');
+var keyDec = (function (){
+    var dict = {},
+        i = keyEnc.length;
     while(i--){
-      dict[keyStrEnc[i]] = i;
+      dict[keyEnc[i]] = i;
     }
     return dict;
   })();
@@ -30,10 +30,10 @@ var Base64String = {
         current,
         status = 0;
 
-    input = this.compress(input);
+    input = this.compressToArray(input);
 
     for (i=0 ; i<input.length ; i++) {
-      c = input.charCodeAt(i);
+      c = input[i].charCodeAt();
       switch (status++) {
         case 0:
           output.push(f((c >> 1)+32));
@@ -181,22 +181,22 @@ var Base64String = {
       i++;
     }
 
-    return this.decompress(output.join(''));
-    //return output;
+    return this.decompressFromArray(output);
 
   },
 
   decompress : function (input) {
-    var output = [];
-    var chr1, chr2, chr3, enc1, enc2, enc3, enc4;
-    var i = 1;
-    var odd = input.charCodeAt(0) >> 8;
+    var output = [],
+        chrc, chr1, chr2, chr3,
+        i = 1,
+        odd = input.charCodeAt(0) >> 8;
 
     while (i < input.length*2 && (i < input.length*2-1 || odd==0)) {
 
       if (i%2==0) {
-        chr1 = input.charCodeAt(i/2) >> 8;
-        chr2 = input.charCodeAt(i/2) & 255;
+        chrc = input.charCodeAt(i/2);
+        chr1 = chrc >> 8;
+        chr2 = chrc & 255;
         if (i/2+1 < input.length)
           chr3 = input.charCodeAt(i/2+1) >> 8;
         else
@@ -204,34 +204,76 @@ var Base64String = {
       } else {
         chr1 = input.charCodeAt((i-1)/2) & 255;
         if ((i+1)/2 < input.length) {
-          chr2 = input.charCodeAt((i+1)/2) >> 8;
-          chr3 = input.charCodeAt((i+1)/2) & 255;
+          chrc = input.charCodeAt((i+1)/2);
+          chr2 = chrc >> 8;
+          chr3 = chrc & 255;
         } else
           chr2=chr3=NaN;
       }
       i+=3;
 
-      enc1 = chr1 >> 2;
-      enc2 = ((chr1 & 3) << 4) | (chr2 >> 4);
-      enc3 = ((chr2 & 15) << 2) | (chr3 >> 6);
-      enc4 = chr3 & 63;
-
-      if (isNaN(chr2) || (i==input.length*2+1 && odd)) {
-        enc3 = enc4 = 64;
-      } else if (isNaN(chr3) || (i==input.length*2 && odd)) {
-        enc4 = 64;
-      }
-
-      output.push(keyStrDec[enc1]);
-      output.push(keyStrDec[enc2]);
-      output.push(keyStrDec[enc3]);
-      output.push(keyStrDec[enc4]);
+      this._chrToOutput(i, input.length, chr1, chr2, chr3, output);
     }
 
     return output.join('');
   },
 
+  decompressFromArray : function (input) {
+    var output = [],
+        chrc, chr1, chr2, chr3,
+        i = 1,
+        odd = input[0].charCodeAt() >> 8;
+
+    while (i < input.length*2 && (i < input.length*2-1 || odd==0)) {
+
+      if (i%2==0) {
+        chrc = input[i/2].charCodeAt();
+        chr1 = chrc >> 8;
+        chr2 = chrc & 255;
+        if (i/2+1 < input.length)
+          chr3 = input[i/2+1].charCodeAt() >> 8;
+        else
+          chr3 = NaN;
+      } else {
+        chr1 = input[(i-1)/2].charCodeAt() & 255;
+        if ((i+1)/2 < input.length) {
+          chrc = input[(i+1)/2].charCodeAt()
+          chr2 = chrc >> 8;
+          chr3 = chrc & 255;
+        } else
+          chr2=chr3=NaN;
+      }
+      i+=3;
+
+      this._chrToOutput(i, input.length, chr1, chr2, chr3, output);
+    }
+
+    return output.join('');
+  },
+
+  // private function to reduce code duplication between
+  // decompress and decompressFromArray
+  _chrToOutput : function (i, length, chr1, chr2, chr3, output) {
+      var enc1 = chr1 >> 2,
+          enc2 = ((chr1 & 3) << 4) | (chr2 >> 4),
+          enc3 = ((chr2 & 15) << 2) | (chr3 >> 6),
+          enc4 = chr3 & 63;
+      if (isNaN(chr2) || (i==length*2+1 && odd)) {
+        enc3 = enc4 = 64;
+      } else if (isNaN(chr3) || (i==length*2 && odd)) {
+        enc4 = 64;
+      }
+      output.push(keyDec[enc1]);
+      output.push(keyDec[enc2]);
+      output.push(keyDec[enc3]);
+      output.push(keyDec[enc4]);
+  },
+
   compress : function (input) {
+    return this.compressToArray(input).join('');
+  },
+
+  compressToArray : function (input) {
     var output = [],
         ol = 1,
         output_,
@@ -243,10 +285,10 @@ var Base64String = {
 
     while (i < input.length) {
 
-      enc1 = keyStrEnc[input.charAt(i++)];
-      enc2 = keyStrEnc[input.charAt(i++)];
-      enc3 = keyStrEnc[input.charAt(i++)];
-      enc4 = keyStrEnc[input.charAt(i++)];
+      enc1 = keyEnc[input.charAt(i++)];
+      enc2 = keyEnc[input.charAt(i++)];
+      enc3 = keyEnc[input.charAt(i++)];
+      enc4 = keyEnc[input.charAt(i++)];
 
       chr1 = (enc1 << 2) | (enc2 >> 4);
       chr2 = ((enc2 & 15) << 4) | (enc3 >> 2);
@@ -282,23 +324,11 @@ var Base64String = {
 
     if (flush) {
       output.push(f(output_));
-      output = output.join('');
-      output = f(output.charCodeAt(0)|256) + output.substring(1);
-    } else {
-      output = output.join('');
+      output[0] = f(output[0].charCodeAt()|256);
     }
-
     return output;
-
   },
 
-  compressToArray : function (input) {
-
-  },
-
-  decompressToArray : function (input) {
-
-  },
 }
 return Base64String;
 })()
--- a/libs/lz-string.js
+++ b/libs/lz-string.js
@@ -15,14 +15,14 @@ var keyStrBase64 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz01234567
 var keyStrUriSafe = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+-$";
 var baseReverseDic = {};
 
-function getBaseValue(alphabet, character) {
+function getReverseDic(alphabet){
   if (!baseReverseDic[alphabet]) {
     baseReverseDic[alphabet] = {};
     for (var i=0 ; i<alphabet.length ; i++) {
       baseReverseDic[alphabet][alphabet.charAt(i)] = i;
     }
   }
-  return baseReverseDic[alphabet][character];
+  return baseReverseDic[alphabet];
 }
 
 var LZString = {
@@ -41,7 +41,8 @@ var LZString = {
   decompressFromBase64 : function (input) {
     if (input == null) return "";
     if (input == "") return null;
-    return LZString._decompress(input.length, 32, function(index) { return getBaseValue(keyStrBase64, input.charAt(index)); });
+    var reverseDict = getReverseDic(keyStrBase64);
+    return LZString._decompress(input.length, 32, function(index) { return reverseDict[input.charAt(index)]; });
   },
 
   compressToUTF16 : function (input) {
@@ -100,7 +101,8 @@ var LZString = {
     if (input == null) return "";
     if (input == "") return null;
     input = input.replace(/ /g, "+");
-    return LZString._decompress(input.length, 32, function(index) { return getBaseValue(keyStrUriSafe, input.charAt(index)); });
+    var reverseDict = getReverseDic(keyStrUriSafe);
+    return LZString._decompress(input.length, 32, function(index) { return reverseDict[input.charAt(index)]; });
   },
 
   compress: function (uncompressed) {
--- a/libs/lz-string.js
+++ b/libs/lz-string.js
@@ -75,20 +75,10 @@ var LZString = {
   decompressFromUint8Array:function (compressed) {
     if (compressed===null || compressed===undefined){
         return LZString.decompressFromArray(compressed);
-    } else {
-        var buf=new Array(compressed.length/2); // 2 bytes per character
-        for (var i=0, TotalLen=buf.length; i<TotalLen; i++) {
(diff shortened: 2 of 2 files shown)


JobLeonard:
Some benchmarks:

plain compress/decompresss:
http://jsbench.github.io/#2a07fd14b55d44291da4b06d3ba6e5c3

base64:
http://jsbench.github.io/#da32b5d1100c24c7744bb58ca3fff440

UTF16:
http://jsbench.github.io/#4029cae03b1e2fc06ad44e35f5bfca6b

Uri:
http://jsbench.github.io/#54c40822dabdbd46a93fb0b7ff6832d9

Uint8:
http://jsbench.github.io/#14395aeb1f452cefa52b5f86b0f644c5

I'd say performance is.. probably a bit better, but also almost negligible within the margin or error.

Given the changes to the code, there _must_ be less allocation going on, so that must have improved a bit at least.

I had them all in one big benchmark, but then I tried to profile them using DevTools. Those tools then suggested that `compress` and `decompress` were deoptimised! The reason, I think, is because I used *all* functions, leading to a different closure being passed each time, leading to de-opting the `compress` function so often that V8 was like "screw this!". So I split it into different benchmarks hoping to mitigate that, since I figure real-life usage doesn't usually use all functions, and because it's unclear when the de-opt hits, making the benchmark less reliable.

Please check if the input data is relevant - I basically took the test string (the one about tattoo's) and tripled it, causing a lot of redundancy.

Rycochet:
I was recently optimising some startup code in an app, and found that the GC could actually have significant impact on things (up to 200ms for 16mb) - but as it's not something easily forced it's hard to get consistent profiling to include that unless running a significantly longer test including `setImmediate()` style callbacks and known consistent memory allocation and freeing.

TL/DR: Reducing the need for GC is always good ;-)

JobLeonard:
The main issue is that it's not easy to see if it's significant in the larger scheme of things - perhaps all computation and memory allocation happens in the core algorithm that I didn't touch.

I think I should set up a plain website with this code, run it through both functions (let's just use the plain compress/decompress as a baseline) with a yuuuge string repeatedly, while running the browser performance tools to keep track of allocations and such. That might give a better overview.

tophf:
Deoptimization is triggered by various patterns (some links: [Deoptimization in V8](https://docs.google.com/presentation/d/1Z6oCocRASCfTqGq1GCo1jbULDGS-w-nzxkbVF7Up0u0), [Optimization-killers](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers)). In your tests, `compress` invocations supposedly use the same parameter signature/types so the actual problem is within the function itself. In my experience, even if there are no obvious deopt triggers in the code, it may be caused by the sheer size of the function and splitting solved that for me. Preventing deoptimization is very important as its impact often/usually negates any perf gains, but unfortunately I'm not an expert either.

JobLeonard:
Hmm, time to dive into this a little deeper and see if we can fix that. Still on sick leave anyway so this is a fun side-track :)

JobLeonard:
The thing I was referring to is that compress is passed a function, which itself is usually a closure. I suspect that level of dynamism prevents some deeper optimisations.

This is pretty old, so probably somewhat outdated, but I'll look into it later:

http://mrale.ph/blog/2012/09/23/grokking-v8-closures-for-fun.html

tophf:
Ah, indeed, no deopt on any of the functions if I test in devtools manually via copypaste, a primitive 100-rep loop, and console.time. The difference is negligible just the same. FWIW when switching to ES6 Map and Set in lz-string I clearly see the advertised 2x gain.

JobLeonard:
Related to that, I just realised there's another optimisation possible for Base64 and URI-safe compressors: don't use `charAt`, use `charCodeAt`. The reason is that looking up integers in hashmaps is much faster than strings. I have some test code below:

```js

// SETUP
var base64 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=";

var charDict = (function (){
    var dict = {};
    for (var i = 0; i < base64.length; i++){
        dict[base64.charAt(i)] = i;
    }
    return dict;
})();

var charCodeDict = (function () {
    var dict = {};
    for (var i = 0; i < base64.length; i++){
        dict[base64.charCodeAt(i)] = i;
    }
    return dict;
})();

var randomString = (function (){
    var i = 100000, strArray = new Array(i);
    while(i--){
        strArray[i] = base64.charAt(Math.random(base64.length)|0);
    }
    return strArray.join('');
})();

var converted = 0;

//BENCHMARKS

// charDict
for (var i = 0; i < randomString.length; i++){
    converted = charDict[randomString.charAt(i)];
}

// charCodeDict
for (var i = 0; i < randomString.length; i++){
    converted = charCodeDict[randomString.charCodeAt(i)];
}
```
http://jsbench.github.io/#a5c234621b81cd41b26e31d7f92f62d4

On my machines integer key lookups are almost 3x faster.

Actually, I think I'll look up the LZW algorithm, read the current one, and see if there are more places where we can use `charCodeAt` instead of `charAt`.

JobLeonard:
@tophf: can you show me what the faster `Map` and `Set` code looks like? Because when I added `Map` to my previous benchmarks it turned out much slower than even plain object lookup there:

- [`char` vs `charCode` combined with `object[key]` vs `Map.get(key)` lookup](http://jsbench.github.io/#f1f9afdba2eac615645c80335c48ed10)
- [`char` vs `charCode` combined with `object.hasOwnProperty` vs `object[key] !== undefined` vs `Map.has(key)`](http://jsbench.github.io/#8345886c28796a6d825c292900d01777)

Anyway, small correction: for Basae64 and URI _decompression_ this is a drop-in replacement.

So looking into this a bit further, there are two different ways we can apply this change:

## Returning an array of `charCode`s instead of `char`s

This _might_ be an optimisation, it _might_ not be. Essentially, all we need to know is whether [`String.fromCharCode.apply(null, charCodeArray))`](https://stackoverflow.com/a/9936506/3211791) is significantly slower than `charArray.join('')`.

If not, we can apply this optimisation safely, essentially replacing my `compressToArray` code with `compressToCharCodeArray`. Also, this would simplify `compressToUint8Array` even further:

```js
  compressToUint8Array: function (uncompressed) {
    //var compressed = LZString.compressToArray(uncompressed);
    var compressed = LZString.compressToCharCodeArray(uncompressed);
    var buf=new Uint8Array(compressed.length*2); // 2 bytes per character

    for (var i=0, TotalLen=compressed.length; i<TotalLen; i++) {
      //var current_value = compressed[i].charCodeAt(0);
      var current_value = compressed[i];
      buf[i*2] = current_value >>> 8;
      buf[i*2+1] = current_value % 256;
    }
    return buf;
  },
```

I'll write a benchmark to test this later, and if it does not immediately make me think it's worse I'll implement it, and benchmark it again.

## Replace the LZW `dictionary` with a trie
Oh goodness, if this works like I _think_ it works, this could be big!

This applies to `_compress`, or `_compressToArray` in my branch.

So this requires a rudimentary understanding of the actual LZW algorithm. I'm using this pseudocode explanation as a guide:

https://www.cs.duke.edu/csed/curious/compression/lzw.html

Basically, LZW is about building a symbol table of prefixes (a JS object), and replacing those prefixes with smaller values in the output stream (an array of integers that then is converted into chars). In a "skipping the crucial-but-currently-not-relevant details" way:

- scan through the input string
- keep track of the current prefix through `context_w`, which starts as an empty string
- store the current character as `context_c`
- `Object.prototype.hasOwnProperty.call(context_dictionary,context_w + context_c)`?
  - `true`
    1. `context_w = context_w + context_c`
    2. `continue`
  - `false`
    1. write the value associated with key `context_w` to the output array
    2. add `context_w + context_c` as a new prefix to the symbol table
    3. `context_w = context_c `
    4. `continue`

I'm skipping a lot stuff - there's a reason the `_compress` function is over 200 lines of code! But this optimisation doesn't apply to it.

So `charAt` comes into play at the `context_c` level. Can we replace it with `charCodeAt`? Yes, I think we can! To do so, we replace the flat dictionary object with a [trie](https://en.wikipedia.org/wiki/Trie) which has nodes that use `charCodes` instead of `chars` as keys:

- scan through the input string
- keep track of the current prefix through `context_node`, which starts at the root of the dictionary, initialised as `{}`
- store the current `charCode` as `context_c`
- `context_node[context_c] !== undefined`?
  - `true`
    1. `context_node = context_node[context_c]`
    2. `continue`
  - `false`
    1. write `context_node[-1]` to the output array
    2. `context_node[context_c] = { val: <new prefix val> }`
    3. `context_node = root`
    4. `continue`

So each node will just consist of numerical `charCode` keys, plus one `-1` key¹ to store the actual integer value of the prefix.

The gain here consists of two things:

- no string appending, so we should have no unnecessary object allocation.
- integer keys instead of string keys, which as the benchmarks I linked above demonstrate are _much_ faster.

And it gets better! Right now there are a lot of calls to `Object.prototype.hasOwnProperty.call()`. I _presume_ this is to avoid the risk of accidentally matching one of the keys of [Object.prototype](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object):

image

In the new version _we'll never use strings for our keys, so there will never be any such collisions!_ Which means we can safely remove all of these calls and just use `object[charCodeAt] !== undefined` instead, which, as you can see in the second benchmark I linked at the top, blows everything else out of the water!

¹ _If we use a string key we lose all benefits of using only numerical lookups - I tested this in benchmarks. Since `charCodeAt` is guaranteed to return a number between 0 and 65535 that means we can use either -1 or 65336. Using 65336 makes the hasing _very_ slow, for whatever reason, but -1 has no negative effect._

Anyway, just writing this down took quite a bit of energy, but now I have an idea of what to do. I'll start working on this tomorrow and see if it turns out as good as I hope it will!

tophf:
> can you show me what the faster Map and Set code looks like? 

https://gist.github.com/tophf/e8962e43efe35233212cf04d8d7cd317

2x speedup compared to the nonminified original version as tested in modern versions of Chrome. Didn't test in other browsers.

JobLeonard:
Thanks, pretty straightforward.

I guess the reason they're not faster in my benchmarks is because I'm limiting myself to 64 single-character keys. Still, the `charCodeAt()` optimisations are kind of orthogonal to `Map` and `Set` so we could even apply both if necessary!

(I wonder @pieroxy is on holiday or something, and will come back thinking "I look away for one second and then this happens?")

JobLeonard:
> The stack is still used for the arguments passed via .apply so anything larger than 32k is a risk depending on the browser. And even that is a risk so I usually do in 1k chunks, then join them.

Ooof... that sounds like performance would fluctuate _wildly_ among browsers and hardware. I'll focus on the other possible optimisations first.

JobLeonard:
Some more thoughts on that de-opting:

```js
let factory = function(){
    return function(){
        return 0;
    }
};
let a = factory();
let b = factory();
let c = a;

a === b; //false
a === c; //true
```

Right now we pass _new_ functions on every call to any compressor/decompressor. If we hoist that we can maybe make things easier for the JIT compilers. However, current set-up requires it because it's `getCharFromInt` and  `getNextValue` close over the passed `input` or `compressed` strings:

```js
  //decompress from uint8array (UCS-2 big endian format)
  decompressFromUint8Array:function (compressed) {
    if (compressed===null || compressed===undefined){
        return LZString.decompressFromArray(compressed);
    } else if (compressed.length == 0){
      return null;
    }
    return LZString._decompress(compressed.length, 128, function (index) { return compressed[index]; });
  },
```

That would require a bit of rewriting too.

pieroxy:
First of all thanks for all the work here. I am eager to see the end result :-)

Just as a side note, be careful when you advertise a x2 increase. It may be so in your browser (lets say Chrome) but then the same code might actually be slower in I.E. or Firefox. When I did a pass at perf optimisations back in the days I probably created 25 jsperf benchmarks and had many surprises on that front. 

That said, all this looks promising.

JobLeonard:
> Just as a side note, be careful when you advertise a x2 increase. 

Right, I should have been more clear that I when I talked about 2x/3x speed-ups was _only_ referring to the micro-benchmarks on (effectively) string vs integer lookup, which is known to be a relatively recent optimisation in browser engines. It doesn't reflect on the whole algorithm or performance across all browsers.

OTOH, the expected lower memory overhead of removing string concatenation (`context_wc = context_w + contextc`) from the core algorithm is probably a bigger deal for old browsers, [since browsers didn't really optimise internal string representation until a few years back](https://blog.mozilla.org/javascript/2014/07/21/slimmer-and-faster-javascript-strings-in-firefox/). So the impact of that in terms of reduced memory overhead _should_ be bigger for older browsers.

At the moment I'm still trying to decipher what you do in the algorithm - specifically lines 137 - 204. It's barely commented, so I only have [your high-level description from the main page to go by](http://pieroxy.net/blog/pages/lz-string/index.html#inline_menu_2) 

> - I initialize the dictionary with three tokens:
>   - An entry that produces a 16-bit token.
>   - An entry that produces an 8-bit token, because most of what I will store is in the `iso-latin-1` space, meaning tokens below 256.
>   - An entry that mark the end of the stream.
> - The output is processed by a bit stream that stores effectively 16 bits per character in the output string.
> - Each token is stored with just as many bits that are needed according to the size of the dictionary. Hence, the first token takes 2 bits, the second to 7th three bits, etc....

(once it clicks for me I'm going to add a modified version of the paragraph below that a documentation comments in the code for later maintainers)

JobLeonard:
At least the bugs are funny:

image

JobLeonard:
So I figured out a simple way to do in-browser line-by-line performance testing: inline `lz-string.js` into `SpecRunner.html`, open the latter, then use the browser dev tool to measure performance for x refreshes. Important: the numbers in the screenshots should be compared only **relative to the other numbers in the same screenshot**, because I'm not refreshing exactly the same number of times!***

First bottleneck I found: -1 and -2 indexing is slow after all:

image

Then I realised "Hey, why not just use a +2 offset on `charCode` for the lookup?"

image

Again: I didn't refresh the same number of times, so you can't compare screenshots directly. Now, `new_node[0] = dictSize++` might look like it's still pretty high, but I think that's the creation of the bucket for the hashtable, which explains why the line after is so fast in comparison.

So then I tried the latest version vs the old version on JSBench, with the profiler on to measure memory use, and using a tripled 'During tattooing...' test string from the tests. Can you spot the point where it goes from old to new in graph?

image

Anyway, moving on to the `_decompressor` next, and then once that is done, check if we can simplify the code (I think going from a flat dictionary to a trie might lead to mismatches in "code quirkiness")

JobLeonard:
So just to add to pieroxy's remark that this is complicated, I did some in-between benchmarking on Chrome and Firefox. I'm using all the test strings except Hello World, so:

- tattoo description (real world text)
- 1000 random floating point nrs (long string with some repetition)
- 'aaaaabaaaaacaaaaadaaaaaeaaaaa' (short string with lots of repetition)
- 'aaaaabaaaaacaaaaadaaaaaeaaaaa' times 1024 (a long string with lots of repetition)
- all printable UTF16 (represents a worst case: long string, no repetition)

On Chrome, the picture is complicated: the new code is slower in quite a few situations, but for long strings with some repetition (1000 random floats) or lots of repetition ('aaa..') it's faster.

On Firefox the picture is a lot rosier, not to mention generally a lot more performant: it's _always_ better, and gets better as strings have more repetition and get longer.

The UTF16 example is actually really revealing, since it is mostly about overhead of filling up a dictionary.

I really hope I can optimise it further so it has either zero regressions in Chrome, or only in insignificant contexts (like really short strings).

image

image

JobLeonard:
I hope my long comments don't feel like spam - I code better if I write out my thought processes, plus I think it helps clarify what I'm changing and why. Apologies for turning this into a micro-blog :P

# **TL;DR: No performance improvements, but I shaved off over half a KB from the minified file**

So the previously discussed optimisations don't apply to decompress. In fact, I don't really see any obvious ways to get it more efficient: it already uses arrays and everything (see profiling below).

However, I managed to get rid of a whole lot of switch statements and even a bunch of variables altogether by applying bit twiddling magic. Frustratingly, the benchmarks all have an error margin of 25% to 50% so it's really impossible to tell what effect it has on performance. It shouldn't be any slower - shifts instead of multiplies and such (again, see below). 

However, there is quite a bit of prefix/postfic increment/decrement abuse going on  in the new code. So I'll add explanations as a comment later. It allows for better minification though: comparing the output from [JSCompress](https://jscompress.com/), the function body goes down from 1304 to 578  characters (in fact, `compress` is a bit shorter too - if I throw the whole thing through a minifier the code is 4023 characters, down from 4823)

An example of the changes:
```js
    // BEFORE
    switch (next = bits) {
      case 0:
          bits = 0;
          maxpower = Math.pow(2,8);
          power=1;
          while (power!=maxpower) {
            resb = data.val & data.position;
            data.position >>= 1;
            if (data.position == 0) {
              data.position = resetValue;
              data.val = getNextValue(data.index++);
            }
            bits |= (resb>0 ? 1 : 0) * power;
            power <<= 1;
          }
        c = f(bits);
        break;
      case 1:
          bits = 0;
          maxpower = Math.pow(2,16);
          power=1;
          while (power!=maxpower) {
            resb = data.val & data.position;
            data.position >>= 1;
            if (data.position == 0) {
              data.position = resetValue;
              data.val = getNextValue(data.index++);
            }
            bits |= (resb>0 ? 1 : 0) * power;
            power <<= 1;
          }
        c = f(bits);
        break;
      case 2:
        return "";
    }
    dictionary[3] = c;
    w = c;
    result.push(c);
```

```js
      // AFTER

      // done once at the top of compress 
      resetValue = Math.log2(resetValue)

      // 0 or 1 implies new character token
      if ((bits&1) == bits){
        maxpower = (8+8*bits);
        bits = power=0;
        while (power!=maxpower) {
          //TODO: Annotate this bit of black magic
          //TL;DR: read out reversed bit-stream, chopped up into chunks of
          // resetvalue-size, but which actually has adaptive-sized tokens.
          // The combination of the two is why reading out is so complicated.
          bits += ((data_val >> --data_position)&1) << power++;
          if (data_position == 0) {
            data_position = resetValue;
            data_val = getNextValue(data_index++);
          }
        }
        dictionary[dictSize] = f(bits);
        bits = dictSize++;
        if (--enlargeIn == 0) {
          enlargeIn = Math.pow(2, numBits++);
        }
      } else if (bits == 2){
        // end of stream token
        return result.join('');
      }
```

### Profiling bottlenecks

I did line level profiling in Chromium, using the aforementioned trick of using the spec (Firefox sadly does not have line level profiling yet AFAIK). I used the raw short repetitive string test and the raw long list of random numbers test, to examine two extremes:

**`aaaaabaaaaacaaaaadaaaaaeaaaaa`:**

short string decompress bottleneck

**1000 random floating point nrs:**

long random nr string bottleneck

As you can see, for short strings creating new chars and adding new tokens dominates. For long strings this goes away. In either case, what the profiling suggests is that our main bottlenecks at this point are:

- additions to the dictionary (unavoidable)
- for strings with many unique chars compared to length: bits-to-char conversion (unavoidable)
- branch mispredictions (_maybe_ we can do a _little_ bit about that, but I doubt it)

My thoughts on that last point: the only way to improve this is loop unrolling. However, that is really complicated here, because we're reading in bits from a bit-stream, in chunks of `resetValue size`, one call to `getNextValue()` at a time, and then processing those bits by `numBits` at a time (which grows as we scan in more characters and prefixes).

```js
        while (power!=maxpower) {
          bits += ((data_val >> --data_position)&1) << power++;
          if (data_position == 0) {
            data_position = resetValue;
            data_val = getNextValue(data_index++);
          }
        }
```

One thing we could try is calling `getNextValue()` multiple times and shifting it into data_val (at most, `getnextvalue` returns 16 bits worth of data, so we can always shift two values in). But because of these out of sync overlapping boundaries of `resetValue` and `numBits` this would probably get very complicated (which is probably what messes up the branch predictor in the first place).

JobLeonard:
So I've looked at `compress` again, also Line Profiling it, using `aaaaabaaaaacaaaaadaaaaaeaaaaa` as the test string. I'd like to zoom in on one particular part (for completeness, [here is the full output](https://user-images.githubusercontent.com/259840/27772436-f96598bc-5f62-11e7-9d24-5dbecffac2c3.png))

image

image

Dictionary assignment is biggest bottleneck (not very surprisingly). Some of that is unavoidable, but we might be able to tweak it. Here's what doesn't work: packing `node[0]` and `node[1]` into an array or object:

```js
            // pack it in an array
            new_node[0] = [dictSize++, c];

            // pack it in an object
            new_node[0] = {v: dictSize++, c: c};
```

I've already tried it (thinking "maybe the problem is that I'm assigning both numbers and objects to the dictionary"), and it's just as slow or slower. So I think we're stuck with that.

That leaves the other dictionary: `dictionaryToCreate`. I think we can actually get rid of that one altogether and replace it with a boolean if we re-organise the order of the algorithm. Right now, what happens is:

> // IN LOOP
> - is `dictionary[charCode]` defined?
>   - false: add it and set `dictionaryToCreate[charCode]` to true
> - does `node[charCode]` exist?
>   - true: `node = node[charCode]`
>   - false: 
>     - Is `dictionaryToCreate[node[1]]` (the charcode in node) defined?
>       - yes? write out `charCode` token to output stream
>       - no? write out `node[0]` to output stream
>     - add new node to trie at current node
>     - set current node to `dictionary[charCode]`
> 
> // AFTER LOOP, clean up last node:
> - does `node[charCode]` exist?
>   - true: `node = node[charCode]`
>   - false: 
>     - Is `dictionaryToCreate[node[1]]` (the charcode in node) defined?
>       - true: write out `charCode` token to output stream
>       - false: write out `node[0]` to output stream
>     - add new node to trie at current node
>     - set current node to `dictionary[charCode]`

But I think that should be reorganisable to:

> // BEFORE LOOP
> - `inputString.length > 0`?
>   - true: 
>     -  write out new `charCode` token to output stream
>      - add `charCode` node to dictionary
>      - `rootNode = true`
> 
> // IN LOOP, **START AT INDEX 1**
> - does `node[charCode]` exist?
>   - true: `node = node[charCode]`
>   - false: 
>     - `rootNode`?
>       - true: `rootNode = false` (don't insert token - already added `charCode` token)
>       - false: write out `node[0]` to output stream
>     - is `dictionary[charCode]` defined?
>       - false: 
>         -  write out new `charCode` token to output stream
>         - add `charCode` node to dictionary
>         - set `rootNode` to true
>     - add new node to trie at current node, that is: `node[charCode] = { 0: dictionary++ }`
>     - set current node to `dictionary[charCode]`

Apologies for the weird pseudo-code, I'm trying to make sense of a pretty tricky algorithm.

Aside from replacing `dictionaryToCreate` with `rootNode`, we also no longer have to store the `charCode` in any nodes (I don't really see this have any significant effect IRL, but still).

I'm probably missing some subtleties, but I'll figure it out as I go.

JobLeonard:
Goodness, that was hard... I mean, about as hard as I expected, but still. In the end the trickiest part was expanding the number of bits at the right pace, had to write some gnarly `console.log` code keeping track of the bitstream to decipher that:

image

Anyway, here are the new stats, running on my Lenovo T440s, Kubuntu 17.04.

### Chrome:
screenshot_20170704_143948
screenshot_20170704_143849

No matter how hard I try, I can't beat the old short-string performance :/. Same with the all UTF16-test. However, I hope you'll agree it's worth losing a bit of performance on those extreme ends for the benefits gained elsewhere. The tattooing text (1100 characters) is already faster now, and when we start to talk about long, very repetitive strings, the performance increases get bigger (aaaaabaaaaacaaaaadaaaaaeaaaaa*1024)

### Firefox:
screenshot_20170704_143807
screenshot_20170704_143819

Was already performing better, and the last code-change improved things even furths.

### MOBILE:





In all honestly, I don't remember which of these was firefox and which of these was Chrome, but given the improvements in performance I don' think it really matters (but note that `81.07 / 7.16 > 11`, so for the long repetitive strings, one of them had a speed multiplication that went up to eleven).

Other fun stuff: post-[JSCompress](https://jscompress.com/) the new code is 3895 characters, or 3.8 KiB vs the old 4.6 KiB, so that's about 17% smaller. I don't know if JSCompress is backwards compatible like LZ-string aims to be though.

@pieroxy, I think I've done all that I can realistically do. Could you test if this works (and performs better) in all relevant contexts? The code has gone through very drastic changes, so I'm sincerely worried about having broken it in some edge-cases.

JobLeonard:
The benchmarks I used, on JSBench:

http://jsbench.github.io/#e25c445d987295b0114407e457dde9ad

http://jsbench.github.io/#4a2ca3e9e3e9ee544f5b76acc7699ef8

If people could please test this out in their own browsers, that would be very useful.

JobLeonard:
In the infamous words oc Columbo: _just one more thing..._

I think that loop unrolling I mentioned earlier is not quite as hard as anticipated (although still hard, and it's still dubious if it gives significant performance benefits). Also, I _highly_ doubt it would help for compression, but for decompression it might.

First, add a variable `chunks = (32 / resetValue) | 0` (calculated after `resetValue = Math.log2(resetValue)+1)`. Since bit-shifting workds on 32 bit values, this represent the number of chars we can read out in bulk and store in `data_val` at once. For standard compression (16 bit UTF) this isn't much: just two characters at once, but for Base64 or URIsafe (six bits), that's five characters at once. That might improve branch prediction.

The only issues is that reading beyond the string length returns `NaN`. That's easy enough to fix though: `Nan | 0 == 0`, and any proper encoded string has an end-of-stream character so those trailing zeros shouldn't matter (this is the part where my code breaks so my assumptions are probably wrong :P).

Then, whenever we need to read in more `charCode`s for `data_val`, we do that in `chunks`:

```js
//OLD:
while (power != maxpower) {
  // shifting has precedence over bitmasking
  bits += (data_val >> --data_position & 1) << power++;
  if (data_position == 0) {
    data_position = resetValue;
    data_val = getNextValue(data_index++);
  }
}

//NEW
while (power!=maxpower) {
   bits += ((data_val >> --data_position)&1) << power++;
   if (data_position == 0) {
     for(i = chunks; i--;){
        dava_val = (getNextValue(data_index++) | 0) << (i*resetValue);
     }
     data_position = resetValue * chunks;
   }
}
```
So we added a while loop inside the if-statement. Is that an improvement? _I don't know_. The thing is that with this code, `if (data_position == 0)` will be false more often, potentially leading to better branch prediction (especially for 6-bit encodings). OTOH, that loop...

However, we can get even nastier with our hack: our encodings are never more than 16 bits, and never less than 6. So `chunks` will be a value between 2 and 5, and we can use [Duff's Device](https://en.wikipedia.org/wiki/Duff%27s_device) and unroll this loop. Also, we  can modify `getNextValue` to guarantee that out of bounds indices always return 0:

```js
while (power != maxpower) {
  // shifting has precedence over bitmasking
  bits += (data_val >> --data_position & 1) << power++;
  if (data_position == 0) {
    dava_val = 0;
    switch(chunks){
      case 5:
        dava_val |= getNextValue(data_index++) << (4 * resetValue);
      case 4:
        dava_val |= getNextValue(data_index++) << (3 * resetValue);
      case 3:
        dava_val |= getNextValue(data_index++) << (2 * resetValue);
      case 2:
      default:
        dava_val |= getNextValue(data_index++) << (1 * resetValue);
        dava_val |= getNextValue(data_index++);
    }
    data_position = resetValue * chunks;
  }
}
```

Would this be faster? Again, _I don't know_. Needs to be tested.

JobLeonard:
Answer after testing: nope, both are slower. So, sticking to the previous version of reading in characters one at a time.

Also, I realised my benchmarks were flawed: I measured both compression and decompression. However, compression is much slower than decompression, so I wasn't properly comparing decompression performance. 

# TL;DR: the new code is faster for decompression in all of my tests

Note that I'm only testing Linux and Android, so it's not exhaustive! Try benchmarks for yourself here:

JSBenchmarks here:
[***Plain Decompression***](http://jsbench.github.io/#da32b5d1100c24c7744bb58ca3fff440)

[***Base64***](http://jsbench.github.io/#7af663013ca789a1141980fbcec43871)

[***Uint8***](http://jsbench.github.io/#4d328bb9d7c5b83a7a5f11c7778c1d16)


## Explanation + screenshots from runs (includes error margins)

***Chrome:*** consistently slightly faster except for the "all UTF16 characters" test. Gets better for very repetitive strings, being up to 1.5x faster.

- [Raw](https://user-images.githubusercontent.com/259840/27868284-90c8fa92-619c-11e7-970f-5aff57eccf9f.png)
- [Base64](https://user-images.githubusercontent.com/259840/27868383-d67a28cc-619c-11e7-99e6-ed55d94a7872.png)
- [Uint8](https://user-images.githubusercontent.com/259840/27868537-4f4e3d24-619d-11e7-80c1-1644289e0515.png)

***Firefox:*** consistently 1.5x - 2x faster for every benchmark, and also faster than Chrome.

- [Raw](https://user-images.githubusercontent.com/259840/27867923-5b7ca164-619b-11e7-93bd-ab04983566e4.png)
- [Base64](https://user-images.githubusercontent.com/259840/27868166-2d0019dc-619c-11e7-8888-11aa04e6141f.png)
- [Uint8](https://user-images.githubusercontent.com/259840/27868672-bab40f3a-619d-11e7-8cce-c1a68118ad0a.png)

***Mobile***

I'm sorry, trying to take screenshots and stitching them together [is really turning into a pain
...](https://user-images.githubusercontent.com/259840/27871313-937bd508-61a5-11e7-9091-df9d674135ff.png). The graphs above are quite telling though.

With Firefox for Android the new code is running laps around the old version. 

In Chrome for Android it's a bit more complicated. For plain compression the new code seems a little bit faster, but the error margins are _more than 100%_ so no idea what that's supposed to mean (negative ops per second?). For Base64 and Uint8 however the new code is consistently a _lot_ faster, in some cases five times faster!

---

Ok, I think I'm now truly done trying to optimise this... I hope there are no nasty edge-cases in my modifications that make it impossible to add this to the main branch.

There is _one_ more thing that I can think of that could be done, but it would break backwards compatibility: don't reverse the bitsstream. Right now compress and decompress writes/reads bitstream tokens one bit at a time, resulting in reverse bits. However, because we know before how long these tokens with a bit of tweaking, we can also just bitmask and shift, but this would remove the reversing (and thus break backwards compatibility because the output string would be different). Instead of a for-loop, we'd have a single if-statement that checks if the token is on the bit-boundary (actually, this is not _entirely_ true, because in theory the number of bits in the token can exceed the number of bits in the stream (especially likely for the Base64 case), so we'd end up with a kind of while-loop thing going on). 

This could end up being a lot faster, especially with decompression where reading in the bits is one of the bigger bottlenecks, according to the above line profiling.

But my sick leave ended yesterday so I should finish up this side-project :P

JobLeonard:
Last bit of optimising, I swear! I can quit when I want to!

Anyway, the `aaaaabaaaaacaaaaadaaaaaeaaaaa` case is still slower, but the gap has been reduced from 20% down to 5%-10%. I also used browserling to test IE10, which also seems very happy with the new version: performance improvements are comparable to Firefox (2x for short strings up to 6x for long repetitive strings).

JobLeonard:
So remember how I said I can't think of any more ways to improve performance? Well, I came up with one possible improvement, but it doesn't seem to work out..

I'll just write it down in case JS engines change their optimisation patterns in the future (who knows, maybe for LZ-string 2.0 somewhere five years down the line, or if someone rewrites the whole function in a compiles-to-WASM language).

The idea is as follows:

- Generally, arrays are faster than dictionary lookups (even with integer keys)
- On the other hand, making an array that can fit all UTF16 characters would require a length of 65k, so using an array as a character node would create huge memory overhead that would slow things down.
- Solution: split up value into upper- and lower- eight bits, and have a mini-tree of nested arrays, each 256 elements large
  - assumption: if other parts of UTF16 are used, these will mostly be similar values, since alphabets are grouped together, meaning we won't have many extra sub-nodes
  - likely downside: we'll be creating two arrays each time we insert a new charCode or prefix, instead of one object. Insertion might be much slower.
- Additional optimisation: the vast majority of charCodes will be ASCII, meaning < 256. We can inline the first array, putting it in the lower 256 indices, and shift the remaining indices by 256, saving us an array lookup:

```js
    if (charCode < 256){
        node = inlinedArray[charCode];
    } else {
        node = inlinedArray[255 + (charCode>>>8)][charCode&0xFF];
    }
```

Whether or not this is faster depends on a lot of things, like the time saved on the second array not created most of the time, array lookup speed, the overhead of an extra if/else branch, and of course how much input data will actually be ASCII values.

So after some first testing, this does not seem to produce consistent performance improvements. I basically tested two scenarios for look-up performance: trying all UTF charCode indices equally, and one win which I assume ASCII charCodes outnumber other charCodes 9:1 (but each charCode was still accessed at some point). Note that neither scenario is very realistic, because most of the time these trees will be relatively empty and only a few indices will be used. Also note that this doesn't test _node creation_ performance, only look-up performance. For that I have a separate test:

[lookup](http://jsbench.github.io/#f90bf23a39841a3da071c51af082aaab)

[insertion](http://jsbench.github.io/#5a054fcfadc140bff6d563010c9fccc2) (I stopped testing early when it was clear this was much slower)

In short: array creation is universally slower, _and_ insertion is slower than look-up (so probably the bottle-neck to optimise for). Array look-up is faster except in IE, where it's twice slower. Combined with the fact that insertion is the major bottleneck (as demonstrated in the line-by-line performance testing in Chrome), that makes this optimisation a no-go.

I can of course still implement it and test it, since real-world scenarios are different from these micro-benchmarks, but I don't have much hope.

---

Slightly different thing: I realised that the only way we can explain the sometimes massive speed-ups in compression speeds for repetitive long strings, is if the new code somehow improved Big O performance. After thinking about it some more, this is actually pretty obvious: the old data structure created ever longer string keys representing prefixes. This means performance is at the very least linearly dependent on the length of these keys, either for hashing or for string concatenation. Since the length of the prefix keys only gets very long for data that compresses very well, that explains why the new data structure sees the biggest speed-up in that scenario. By comparison, the new data structure only uses integers, meaning it has the same time overhead for every value.

JobLeonard:
So I've been trying if the trie-approach might work for `js-search` as well. At first it seemed to be slower, until I tried an array-based approach:

> 
> - typically, we'll be dealing with a small alphabet (Latin alphabet is 26 letters, times two for capitalisation, plus punctuation and some special characters)
> - of said alphabet, we are probably only using a subset in our words
> - of said subset, _as soon as we're beyond the root of a prefix_, we'll have an even smaller subset
> 
> With that in mind, a linear search through an array is probably going to be faster than any fancy algorithm, the same way that insertion sort beats everything else for arrays less than twenty items or so, through sheer simplicity and cache friendliness.

I just tried if the same structure would perform well with compression. It doesn't:

http://jsbench.github.io/#a868adf796802a51a4e9add318572c70

I'm not even testing "all UTF16", because from the Jasmine test I already know basically freezes the tab - it's worst case behaviour after all, being _at least quadratic_. Another reason not to use this.

JobLeonard:
Ignore my previous claim: jsbench seems to have lots of overhead messing up the benchmarks due to GC pauses or something, because I just tried again on jsperf and the array approach is much for all normal data:

https://jsperf.com/lz-string-improvements-short-strings

image
image




... _however_, that worst case performance (i.e. all UTF16) is bothering me. I'm not really worried about real data: if _random numbers_ still compress faster, I'm sure average performance is better for all normal data people will use LZ-string for. However, that assumes people _won't send in bad data on purpose to ruin your day_. And I don't know how fast performanc degrades - what about compressing Chinese?

So my suggestion is to put this in a separate file, say `lz-string-unsafe.js`, enforcing that using this version is a conscious choice where the programmer looked at the potential issues. Because there are scenarios where you can guarantee it's safe to use and the benefits will be significant (like in my case, I only compress data we generate ourselves - and it's sparse data so the speed improvements will be significant).

ronag:
@JobLeonard Wouldn't using Map with a polyfill fallback further improve performance?

JobLeonard:
Well, it definitely might be, but perhaps it makes more sense to have a separate map version and let people use polyfills in their build system. Otherwise you get code duplication.

Rycochet:
You can often use `Object.create(null)` instead of a `Map` in a large proportion of the times you want to use a `Map` - don't know if that's any use, but worth mentioning :-)

i336:
I found this page via a Hacker News comment (https://news.ycombinator.com/item?id=14983356) and asked in a reply if benchmarks from an old PC would be useful.

This is my ThinkPad T43 - a single-core 32-bit Pentium M, with 2GB RAM - on **the last mentioned* jsPerf link, https://jsperf.com/lz-string-improvements-short-strings. I'd run all the jsPerf links mentioned here, but I'd drown the thread in what would probably be mostly irrelevant screenshots. So, let me know what others may be interesting.

Oh, FWIW, the CPU was clocked at 2GHz, as this laptop is almost always plugged in.

image

koudelka:
It'd be wonderful to see this merged. 👍

pieroxy:
So... I haven't got the time to review all this lengthy thread. Thanks for all the effort though. 

What's your thoughts about merging this? Should I merge everything at once or should I merge bits and pieces ?

I unfortunately don't have the time to review everything in details, but I'll make a quick pass before merging anyways.

JobLeonard:
Hey, sorry for the slow reply.

I'm... not sure? I mean, my version passes all the tests you set up, and it seems to work for everything I throw at it. On the other hand, I changed things pretty darn dramatically so I might have missed an edge-case or two that the tests don't cover. Like someone already having data out there that suddenly doesn't decompress properly. I'm tempted to suggest a version bump, just to indicate that this _might_ break things.

rquadling:
We use LZString and store about 12MB into localStorage - takes about 2MB - With a version bump, I can certainly give some real world testing.

JobLeonard:
Maybe a release candidate-type approach before merging it? I _really_ don't want to risk breaking anything in production! I mean, look at these stats:

> 2,014 downloads in the last day
> 10,353 downloads in the last week
> 43,481 downloads in the last month

https://www.npmjs.com/package/lz-string

That's quite a lot of packages I might break with these code changes...

rquadling:
An RC would be great. Those who pull bleeding edge will get it (and understand they are pulling bleeding edge).

rquadling:
So, 2.0.0-RC1 please!!!

JobLeonard:
It's _supposed_ to be backwards compatible though, so it shouldn't get a major version bump.

Rycochet:
@JobLeonard I'm actually doing something similar in another project - there are going to be a couple of incompatibilities (though not many) - mostly it's just a major rewrite including Javascript -> Typescript, and complete changes to how it's actually doing things internally. If it was only the internal changes then I'd be sticking it as a minor bump, but because of the way it's a total rewrite I'm putting it as a major bump so if there are issues then they're totally separated and on their own version "tree" (it's also got well over 100m users a month - so I can't afford to risk it lol).

Considering the performance benefits, and the complete change in how it's implementing the algorithms, I'd understand putting it as a minor bump, but be a lot more confident with a major bump. At work I make use of the compression, which then gives the decompression side to clients (to save bandwidth etc), so I'm definitely going to be using this (and it also means several hundred thousand more un-tracked users who have no idea it's actually there).

It's totally up to @pieroxy what he wants to do, but my preference is obviously a major bump. It might not be strict semantic versioning as per http://semver.org/ - but it is the "safer" way to do it.

pieroxy:
Ok, 2.0.0-RC1 it's going to be. I'm going to do extensive testing in browsers to ensure the current "release" version and the optimized one do produce the same result.

I will hopefully get some free time in the following weeks so this will move forward soon :-)

bubbatls:
On my set of data, with this version, compression is nearly 3 times faster than the original, so I'm very interested in the merge.

tophf:
The `unsafe` version is impressive when compressing 1MB of json'ified array of 100 CSS stylesheets with LZStringUnsafe.compressToUTF16 - `350ms` vs `850ms` (official) vs `650ms` (this PR's lz-string.js).

JobLeonard:
Yes, it's pretty fast! It basically makes use of two things: `array.push()` is faster than dictionary assignment, and the fact that up to a certain length, linear search over an array is much, much faster than anything else on modern machines.

> - typically, we'll be dealing with a small alphabet (Latin alphabet is 26 letters, times two for capitalisation, plus punctuation and some special characters)
> - of said alphabet, we are probably only using a subset in our words
> - of said subset, _as soon as we're beyond the root of a prefix_, we'll have an even smaller subset
> 
> With that in mind, a linear search through an array is probably going to be faster than any fancy algorithm, the same way that insertion sort beats everything else for arrays less than twenty items or so, through sheer simplicity and cache friendliness.

However, the limitation is the degenerative case of having a huge alphabet. Then linear search, being O(n), slows down to a tab-freezing crawls.

That makes it a risky target for malicious attacks, if people send strings containing all UTF encodable characters. However, if you are in control of the data being compressed, like with your own CSS, it should be fine.

I've actually been thinking of using [a home-made Uint32Vector](https://run.perf.zone/view/Crude-minimal-Uint32Vector-vs-Array-1512389144330) class to implement a trie by faking low-level pointers using the array to store indices to itself. That would avoid this degenerate case. Each trie node would implement its own micro-subtree storing the 16 charCode bits. I did a quick sanity check for look-up performance at different tree depths (eight, four and two):

https://run.perf.zone/view/Typed-Array-based-trie-vs-object-trie-sanity-check-1512394647312

https://run.perf.zone/view/Typed-Array-based-trie-vs-object-trie-sanity-check-2-1512395700764

https://run.perf.zone/view/Typed-Array-based-trie-vs-object-trie-sanity-check-3-1512396492187

The main issue is that this would create a huge memory overhead: each charCode node would require a sub-tree. The eight-level tree is smallest (and slowest) subtree in terms of bytes, at 132 bytes per leaf node (interestingly, this number goes *down* the more filled in the node gets: a node pointing to two characters is 66 bytes per charCode, and if it points to _all_ charCodes it's just barely over four bytes per charCode). 

In simple terms: that's worst-case memory blow-up of 132x the input string. The faster one (two levels) is 516x. Throw in an 8MB file, eat up 4GiB! (in reality that number would be lower, because of the reduction in overhead mentioned above).

Still, even the slow eight-level tree might end up being a lot faster because *assignment* is potentially a lot faster than dictionary assignment, which is the bigger bottle-neck.

JobLeonard:
@tophf, could you share the 1MiB worth of JSONified CSS? Could be a good test-case for performance & reliability.

tophf:
@JobLeonard, I've downloaded some from the [most popular](https://userstyles.org/styles/browse) on userstyles.org: [jsons.zip](https://github.com/pieroxy/lz-string/files/1548952/jsons.zip). If you want more, use `https://userstyles.org/styles/chrome/###.json` link where `###` is the style id (the only number in the style's page URL).

rquadling:
Is this anywhere close to a release please?

JobLeonard:
Ehm.. well.. I was about to create a backwards-compatibility breaking fork (but with theoretically better compression) and I'm more interested in developing that :P

But if there are at least two others who want to join in I'm happy to be one of the maintainers, at least.

Rycochet:
I'd happily be another maintainer - though fair warning, I'm a huge fan of Typescript and would be hoping to convert this over to it at some time (though tbh the only part needed in here is typings, linting and tests). Saying that, I'm also super busy hence only volunteering as "one of many" ;-)

rquadling:
I'd be happy to offer my limited JS skills. I'm mainly a PHP dev, but have some experience with JS. But not to the degree that has been exhibited here. Happy to do releases and coordinate that side of things.

JobLeonard:
Ok, so I've just been made one of the maintainers of this repo, @Rycochet, @rquadling too I guess? :)

How shall we proceed? Does anyone feel like code-reviewing this, or shall we move on to a 2.0 RC and let the bug reports sort it out? :P

pieroxy:
I'd go with the second option. Let's also setup a jsperf so that we can compare both versions on a good range of browsers.

Yes, you are all three maintainers :-)

rquadling:
I'd go with an RC and see what's what.

rquadling:
I'd say that the review of this has been pretty thorough. Getting the code out in the wild in an RC and see what people say about it.

pieroxy:
https://jsperf.com/lzstring1vs2/1

Let the tests begin

pieroxy:
Hmmm, I can't see the results of my various tests... Did they remove this feature?
The old version is 75% slower on FF, I11 and Chrome on Windows. On FF linux it's 55% slower.

JobLeonard:
> var data144 = document.body.innerHTML;

Hah, neat trick! I think we should also create a decent body of realistic data sources to test against though. I already highlighted the pathological case for the `unsafe` method

JobLeonard:
I managed to squeeze more blood from this stone.

## Unminified

photo_2018-07-04_16-18-49

photo_2018-07-04_16-18-47

image

image

The only (minor) regression is desktop Firefox, and that one is already ten times faster than anything else anyway

https://jsperf.com/lzstring1vs2vs2ss

## Minified

The refactored StringStream makes the code smaller and easier to follow, minifies a bit better (about 200 bytes) but compresses slightly worse (1.4kb instead of 1.3kb.. but seriously, who cares about 100 bytes in this day and age?)

Again it's faster in all browsers except the already exceptionally fast Firefox

photo_2018-07-04_16-51-06

photo_2018-07-04_16-51-00

image

image

JobLeonard:
I'll apply these changes to the Unsafe version later, see what happens

rquadling:
It seems you are all approaching that time when you can compress an infinite amount of data in zero time into zero space : ∞ => 0/0

JobLeonard:
Heh, well for a coarse enough definition of approaching I guess :)

(BTW, in retrospect I have no idea why I didn't immediately go with the unsafe version for the Uint8Array version - the maximum length is 256, not 65k, so the degenerate case is not nearly as bad)

JobLeonard:
https://jsperf.com/lzstring-1-vs-unsafe-vs-unsafe-ss/1

Less conclusive this time: faster on Chrome, slower on Firefox desktop *and* mobile (although these are faster than Chrome anyway).

photo_2018-07-04_20-46-54

photo_2018-07-04_20-46-50

image

image

Personally I'm in favour of "optimise for the slowest browsers", which is Chrome in this case. We should test Safari and Edge too though.

Also, I have an idea that might be faster still, although I think it is unlikely to make a difference: instead of a prototypical object, turn the stringstream methods into functions that are called from within `_compress` with the variables that it used to have. Save on prototypical lookup, `this.` dereferencing but adds variables that have to be passed and hence garbage collected. And perhaps the JIT optimiser works better with inlining optimised functions it than it would with prototypical methods.

JobLeonard:
> I have an idea that might be faster still, although I think it is unlikely to make a difference

It's worse than that: it's even slower than the old version!

https://jsperf.com/lzstring-1-vs-unsafe-vs-unsafe-ss/3

JobLeonard:
https://run.perf.zone/view/LZString-variants-11-07-20182018-minified-long-string-of-As-1531314387251

https://run.perf.zone/view/LZString-variants-11-07-20182018-minified-long-string-of-random-1s-and-0s-1531314038065

https://run.perf.zone/view/LZString-variants-11-07-20182018-minified-long-string-of-random-letters-a-zA-Z-1531314312127

https://run.perf.zone/view/LZString-variants-11-07-20182018-minified-compress-documentbodyinnerHTML-1531315228392

gloryknight:
Indeed, as I have suggested. Minified version is much faster (at least in Chrome). Well done.

franciscop:
This progress is amazing! I'm using `lz-string` (actually `lznext`, which is just a thin wrapper for `import`) and the only small disadvantage of `lz-string` is that it's a bit slow. I've added a [~100kb limit for compressing text](https://github.com/franciscop/brownies/blob/master/src/packer.js#L14) (around 60ms) in [my library `brownies`](https://github.com/franciscop/brownies) to avoid blocking the main thread:

```js
export const pack = str => {
  // ...

  // Compress it only for relatively small strings, since compression is O(N)
  //   so it takes too long for large strings
  if (str.length < 100 * 1000) {
    str = lz + LZString.compressToUTF16(str);
  }
  return str;
};
```

JobLeonard:
@franciscop: for the record, [I wrote an async version](https://github.com/pieroxy/lz-string/issues/113#issuecomment-414814677) - it's slower but non-blocking

franciscop:
That's pretty cool. Unfortunately that would imply a full rewrite of my library and changing the API radically, since this would no longer be possible:

```js
import { local } from 'brownies';

local.id = 10;
console.log(local.id);
```

I'm using Proxy internally for that setter, and a setter cannot be awaited... (I could set a local cache and do sync on the background, but that's a totally different can of worms).

paultman:
Any idea when this going to be merged to master and a 2.0 release available?
I'm currently using localstorage and would like to move to localForage and maybe optimize the string values using this compression as well.  I know it would help with localStorage, and imagine it should be the same with the localForage wrapper as well.

anonyco:
Use ASM.JS integer optimizations to further increase speed by a few percentage points. Both Chrome and Firefox have considerations for integer optimizations. Observe the benchmark below.

```Javascript
requestIdleCallback(function(){
    for (var i=0; i<16777216; i=i+1|0) void 0; // warm up

	const console = window.console;
	
	function runWithoutSpeed(){
		console.time("SlowerVersion");
		for (var i=-4194304,j=4194304; i < j; i+=2, j-=2) void i + j;
		console.timeEnd("SlowerVersion");
    }
	function runOptimizedTest() {
		console.time("OptimizedTest");
		for (var i=-4194304,j=4194304; i < j; i=i+2|0, j=j-2|0) void (i+j|0);
		console.timeEnd("OptimizedTest");
    }
	runOptimizedTest();
	runWithoutSpeed();
}, {timeout: 11});
```
To do integer optimizations, you must explicitly cast the number to an integer after every pass via function arguments, every addition/subtraction chain, every division. Also, avoid multiplication wherever possible. Use bit-shifts instead.

Rycochet:
Output in my Chrome -

```
OptimizedTest: 2.701171875ms
SlowerVersion: 3.35595703125ms
```

anonyco:
Output in my Chrome -

```
OptimizedTest: 4.52978515625ms
SlowerVersion: 5.72216796875ms
```

Output in my Firefox:

```
OptimizedTest: 2ms
SlowerVersion: 3ms
```

Indeed, integer optimizations do make Javascript faster.

JobLeonard:
That's not really `asm.js`, but integer optimization (or [SMI optimization](https://ponyfoo.com/articles/javascript-performance-pitfalls-v8#double-fields) if you're on 32 bit V8). Sadly, "proper" `asm.js` [is a bit more involved to write](https://github.com/zbjornson/human-asmjs). It's really hard to do right, and not really practical for our case: we need to initiate our own heap, do memory management, and wait for the asm.js compiler to do its thing (which is pretty slow). The start-up time might be worse than the actually performance boost, and it is very likely that we end up making a mistake and the whole thing turns back into regular JS.

I actually was trying to clean up the PR (locally, on my machine) a month ago, *and* finally got around to implementing an unsafe variant that made use of typed arrays to see if that was even faster. It was:

image

However, in the process I discovered another bug in the new version, and I haven't managed to fix it in the meantime.

anonyco:
"Proper" asm.js code is impossible to write for these purposes because we are working with strings which are foreign to asm.js. Nevertheless. integer optimizations do exist. I am very glad to hear that the Uint32Array version is much faster, but I know that it could be a lot faster if put integer optimizations into both the loops and variables surrounding the Uint32Arrays. If you could please post your LZStringUnsafe32, I could take a look at it for bugs and optimizations.

tajetaje:
I take it that this PR isn't getting merged in anytime soon, but is the code in [JobLeonard/lz-string](https://github.com/JobLeonard/lz-string/tree/array-lookup) safe to download and use directly? Or should I stick with the version on npm/main branch here?

JobLeonard:
Yeah, sorry, the truth is that I'm just terrified of breaking something because I never published a package to NPM before and this is a package that still gets used in tons of places.

But that isn't really about the code. In that regard it passes all existing tests, should be perfectly backwards compatible, and [Stylus](https://addons.mozilla.org/en-US/firefox/addon/styl-us/) uses it in their code-base (or at least used to, I don't know if they still do) and afaik it hasn't given them any trouble.

tajetaje:
> Yeah, sorry, the truth is that I'm just terrified of breaking something because I never published a package to NPM before and this is a package that still gets used in tons of places.
> 
> 
> 
> But that isn't really about the code. In that regard it passes all existing tests, should be perfectly backwards compatible, and [Stylus](https://addons.mozilla.org/en-US/firefox/addon/styl-us/) uses it in their code-base (or at least used to, I don't know if they still do) and afaik it hasn't given them any trouble.

Maybe set it as version 2, make it clear that it has potential to be be a breaking change? Or just publish it under @JobLeonard/lz-string?

rquadling:
Releasing as a major version upgrade should be enough to let everyone know that there MAY be BCs. Semantic versioning and all that.

rquadling:
If everyone is happy for this to be released then I can do that as v2.0.0

rquadling:
With regard to the minification process .. this doesn't seem to be documented anywhere. Should we not be doing that as part of a release process? Normally accepting compressed files without ensuring they are based upon uncompressed ones is a bad idea.

JobLeonard:
I think I just threw it through a minifier, but yeah, that probably should be standardized somehow

rquadling:
There is a PR on here that relates to more pipeline work. Ideally someone with JS skills should come and look at it. I'm not that person. I work as part of the team that uses this library.

Rycochet:
I feel that the release process should be a different PR to this - also that this should be merged, then followed by an almost immediate conversion to Typescript in another PR (straight conversion, no code changes - so not updating #123) - using perhaps https://www.npmjs.com/package/microbundle (I used to use TSDX, but that's been abandoned for a couple of years) - then a further PR to add CI (Travis or Github actions?) test build and release :-)

Kikobeats:
There are a lot of good work on this PR! 

merge this and release a v2 should be enough to don't break anything 🙂

PR #102, opened 2017-09-03
insure the correct format parameters

RicoLiu:
insure the correct format parameters


Rycochet:
Not sure what this was fixing, but the tests pass so if needed please reopen as an issue with data to test against

PR #116, opened 2018-07-09
Optimized code for smaller size and faster execution.

gloryknight:
About 25% smaller size of minified file by replacing Math.pow with a reference variable, replacing data object with variables, constant values are fixed, string of symbols are replicated where possible, repeating block of iteration over power variable and repeating comparison are enclosed into functions. No change to logic or output. Execution speed [ LZString.decompress(LZString.compress(s)) ] is increased by about 20% (depending on payload).


JobLeonard:
Thanks for the suggestions!

Sadly, it looks like your changes are (mostly) superseded by https://github.com/pieroxy/lz-string/pull/98, which is so thoroughly changed it barely resembles the current code base. For example, removes `Math.pow` altogether, and uses an array for `reverseDict`, as that is faster. So it's not immediately obvious if your optimisations still apply, or if a similar change has already been applied by me.

If you're interested, [start from the `array-lookup` branch of my fork](https://github.com/jobleonard/lz-string/tree/array-lookup). If you see any way to improve it, I'd be more than happy to merge a pull request on that branch from you! :)

Issue #122, opened 2018-10-19
Proposal: define an LZString spec + version scheme


JobLeonard:
So there is no official spec at the moment - the JS implementation is the reference implementation. 

Perhaps it is worth making a clear spec instead. Here is my proposal for what to consider when makning (not for a spec itself just yet - after all, let's first agree whether we want one or not).

## High-Level Overview

An LZString spec 1.0 (LZ1) sould unambiguously describe the current formats for each compression variant. My high-level proposal is that we structure it as:

1. The compression/decompression algorithm itself, up until the part where tokens are sent to a bitstream, or decompressed from a bitstream.
2. The different bitstream-to-string and back algorithms, that gives us the many variants that we currently have.
3. Easily understandable, highly commented reference implementations (as in, *not* my fast implementation :P)
4. Unit Tests that fit the spec and edge cases?

Points 1 and 2 basically reflect the decomposition we currently use in the library already. 3 and 4 are good to help others implement their own ports, and probably a good exercise to find any bugs.

The versioning is of the compression _format_, not the library; changing the library version has no effect on the specification of the format as long as it compresses/decompresses the same.

## When To Make A New Spec

LZString spec 2.0 and beyond (LZ2+) has yet to be written, and should be used for the situation where either of the above (or both) would break backwards compatibility.

Now to be clear: just like how I think it's cool that the library is pre-ES5 compatible and would love to maintain that just as a kind of badge of honor despite diminishing needs, supporting LZ1 for as long as possible would be a Very Good Thing in this world of constant backwards-compatibility upgrades.

But that is part of the point: now, it is kind of implied that a new version of LZString will still work. But 

There must be really compelling benefits to create an LZ2+ spec, and it should not happen lightly (I might have some ideas, but that's a story for another topic).

I also think that the parts that break *should break in a reliable way, even with old libraries*. Which leads us to:

## The Backwards Compatible Header Hack

If we go so far as to make a new compression scheme that breaks with LZ1, it would be cool if there were some kind of way for LZString libraries to detect which compression version is being used. Even better would be if we would not have to worry about LZ2+ compressed strings being passed to *old LZString code*. Also, we might want to be able to add more metadata (although I really don't see what kind of metadata that would be TBH).

Well, I think I came up with some possible tricks for all of that! First, note that LZ1 has the following default values for tokens:

0: new 8-bit char code
1: new 16-bit char code
2: termination token

Elsewhere, I suggested that for a hypothetical LZ2+ we re-order that to:

[...]

PR #127, opened 2019-01-29
Fixed safe uri alphabet (+$ -> _.) for compressToEncodedURIComponent/decompressToEncodedURIComponent

r0b-:
I don't know why '+' and '$' were chosen when this feature was implemented, but both characters need to be encoded for valid query strings. (`encodeURIComponent('+$') => '%2B%24'`)
So I changed them to '_' and '.' ... and also removed the +/whitespace replacement as it's not needed anymore.

### DIFF
--- a/libs/lz-string.js
+++ b/libs/lz-string.js
@@ -12,7 +12,7 @@ var LZString = (function() {
 // private property
 var f = String.fromCharCode;
 var keyStrBase64 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=";
-var keyStrUriSafe = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+-$";
+var keyStrUriSafe = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_-.";
 var baseReverseDic = {};
 
 function getBaseValue(alphabet, character) {
@@ -99,7 +99,6 @@ var LZString = {
   decompressFromEncodedURIComponent:function (input) {
     if (input == null) return "";
     if (input == "") return null;
-    input = input.replace(/ /g, "+");
     return LZString._decompress(input.length, 32, function(index) { return getBaseValue(keyStrUriSafe, input.charAt(index)); });
   },
 


JobLeonard:
Thanks, this is a known issue but we cannot merge this unfortunately :(

The problem is that we cannot change this without breaking backwards compatibility.

(also, definitely something to remember if we ever go through a "spec" change - see #122)

pieroxy:
While I agree we should do this, this is a breaking change and will break havoc all over the place. 

So we either reject it or we create another method...

wkrick:
@r0b- @pieroxy

I propose two new methods...

- `compressToURISafe()`
- `decompressFromURISafe()`

With "URI Safe" being the "Unreserved" characters defined in [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt) as discussed here...

https://joshbuchea.com/uri-safe-characters/

...but excluding the tilde character '~' for historical reasons due to [RFC 1738](https://www.ietf.org/rfc/rfc1738.txt) as discussed here...

https://jkorpela.fi/tilde.html

I think that naming the methods the way I suggest above makes it clearer what the encoding is actually doing. As "URI Safe" can be thought of as a character encoding like "UTF-16" so the new method naming should parallel the existing naming (i.e. `compressToUTF16` / `decompressFromUTF16`).  The current method names (which allude to [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) and [decodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent) ) make it sound like the result will have URL-encoded values using percents or something, which isn't really what it does as the character set used by the method (ignoring the current bug) should produce a string that doesn't actually need to be URI-encoded (i.e. using percents) at all.  I totally understand why the original name was chosen, but I think it could be clearer for the user.

Possible alternate (but longer) method names if the ones suggested above aren't acceptable:

- `compressToURISafeEncoding()`  `decompressFromURISafeEncoding()`

- `compressToRFC3986URISafe()`  `decompressFromRFC3986URISafe()`

- `compressToRFC3986Unreserved()`  `decompressFromRFC3986Unreserved()`

- `compressToRFC3986UnreservedURICharacters()`  `decompressFromRFC3986UnreservedURICharacters()`

etc...

wkrick:
@r0b- @Rycochet 

I just came across something called Base64URL which is described in Section 5 of RFC 4648 back in 2006...

https://www.rfc-editor.org/rfc/rfc4648#section-10

...and explained in more detail on this website...

https://base64.guru/standards/base64url

Base64URL is addressing the same problem with regards to Base64 encoded strings containing characters that aren't allowed in URLs.

In short, it's the same algorithm as Base64 except for these 4 changes:

- Replaces “+” by “-” (minus)
- Replaces “/” by “_” (underline)
- Does not require a padding character
- Forbids line separators

I feel that the new compressToURISafe() method (or whatever it ends up being called) should use the same implementation as described in.

PR #137, opened 2019-07-20
Implement tree-shakeable build process, code cleanup

mgdodge:
The update seems large, but is mostly code formatting. Existing functionality is preserved, all unit tests still pass. Even loaded in angularjs to verify umd wrapper correctly registers it when angular is present.

**New features**

- Unit tests migrated from jasmine to jest (which is based on jasmine, tests themselves remained largely untouched!)
- Rollup build process implemented - produces 3 builds: iife (minified for browser use), umd (for legacy apps, angularjs, commonjs, etc), and esm build (for modern bundlers, **_tree-shakeable_**!)
- Eslint/editorconfig implemented for code quality and formatting consistency

**Cleanup Items**

- Convert existing files to proper indentation based on these two changes
  - Code formatting based on modified airbnb standards - tabbed indentation to please @Rycochet [mentioned in this other PR](https://github.com/pieroxy/lz-string/pull/123#issuecomment-436239257)
  - Editorconfig to encourage future updates follow proper style guidelines
- Replace code flagged by eslint with functional equivalent - no logic lost/changed

**Note**: One eslint rule indicated two functions were used before they were defined. While in older versions of JS this was not an issue because of hoisting, [use of const/let in the ESM version](https://eslint.org/docs/rules/no-use-before-define) requires those functions to be defined earlier in the file.


fregante:
I want ES Modules too but to improve the chances of this being merged, it's best to avoid _any_ unrelated changes, which this makes a lot.

This PR should probably be reduced to just `package.json` and `rollup.config.js`

Also `rollup` can generate multiple bundles at once so you can further simplify its config.

>   output: { // required (can be an array, for multiple outputs)
>
> From: https://rollupjs.org/guide/en/#configuration-files

mgdodge:
Understood. Will close this PR and submit new _minimal_ one with just rollup build (slightly more than just package.json is required, but not nearly as many changes as seen here.) Since other changes would be best applied after build process is in place, I will submit those separately after this one is merged.

PR #138, opened 2019-08-01
Implement build process to allow ES module support (ie. tree-shakable)

mgdodge:
All existing functionality is preserved, all unit tests still pass. Even loaded in angularjs to verify umd wrapper correctly registers it when angular is present.

**New features**

- Rollup build process implemented - produces 3 builds: iife (minified for browser use), umd (for legacy apps, angularjs, commonjs, etc), and esm build (for modern bundlers, tree-shakeable!)

New files added outside of `lib` folder are for build process/npm support. Version numbers bumped to 1.5.0 in package.json/bower.json - when published to npm, these should be in sync. Retain copy of 1.4.4 in `reference` folder.

(This PR is a subset of changes from #137, focusing on just the build process. Remaining changes from that PR will be submitted separately.)


mgdodge:
Any chance of this getting a response?

rquadling:
Hi. Really sorry there's been no response. Unfortunately I cannot comment on the actual PR, but the principle here is good. I hope that others will be able to take a better look.

sukima:
The produced output looks correct to me. What are the chances of getting this merged?

Rycochet:
Needs someone with some time to look at it - the project needs a bit of updating regardless (ie, bower was deprecated ~5 years ago, should be in typescript, and there's another PR with some nice performance updates) - it's not large enough to be a major issue for most people so nobody really puts this high on their priorities for checking (the downside of a project that does its job well!)

Rycochet:
The new version uses `vite` and `rollup` to ensure there's both a real esmodule and commonjs build available!

PR #140, opened 2019-11-13
Allow to specify the function to use as a CLI command arg

Lucas-C:


Lucas-C:
Any feedback on this @pieroxy ? :)

README.md
```diff
@@ -16,6 +16,7 @@ If you are using one of the ports of lz-string to decode on the server what was
 ```shell
 $ npm install -g lz-string
 $ lz-string input.js > output.txt
```
rquadling, line 18: My preference would be to have `compress` and `decompress` as a pair of "commands".

The group seems to be ready to release a potentially BC in a new version, so having the CLI also have a BC is not a no-no.

bin/bin.js
```diff
@@ -3,11 +3,18 @@ var lzString = require('../libs/lz-string.js');
...
+  console.error('Usage: lz-string [funcName] <input_file>');
   process.exit(1);
 }
 
-console.log(lzString.compress(fs.readFileSync(process.argv[2], {
+var funcName = 'compress',
+    filePath = process.argv[2];
+
+if (process.argv.length == 4) {
+    funcName = process.argv[2];
```
rquadling, line 14: This would/could effectively expose any method available. If we go with having `compress` and `decompress` as the "command" (in `process.argv[2]`), then that may make this a little more consistent overall.

Rycochet:
The CLI function is being totally rewritten, with support for all functions

PR #145, opened 2020-05-10
Added support decompressFromUint8Array to save 1 bytes on decoding

mochouinard:
Other implementation of LZString doesn't always send the last byte if it 0.
This cause issues with the decompressFromUint8Array which devide the
length by 2 which would return an float.  So I rectify it to it ceiling
value.


mochouinard:
Well, I need to rework this... It doesn't work... I do need to append the 0 at the end... Will resubmit when it work better !

PR #152, opened 2021-04-21
Chrome extension

hbazan:
Initialization for Chrome Extension contexts


libs/lz-string.js
```diff
@@ -496,6 +496,8 @@ var LZString = {
 
 if (typeof define === 'function' && define.amd) {
   define(function () { return LZString; });
+} else if (typeof chrome !== "undefined" && typeof chrome.extension !== "undefined") {
```
hbazan, line 499: if we are running on chrome and on a chrome extension context, the scope is the "UnsafeWindow" where the extension is running

pieroxy:
@Rycochet @JobLeonard @rquadling Any objection to merge this ?

JobLeonard:
LGTM

Apologies, @hbazan, I had overlooked this in my notifications in April

phanirithvij:
I believe this can be merged @pieroxy.

rquadling:
I'm only 21 months late! Sorry. I'm not a JS expert, but can help deal with releasing if that's a requirement still.

Rycochet:
Initialisation has changed and is now build via `vite` and `rollup` - I also try to use `globalThis` instead of window etc, so can you check the rc builds (not currently released to npmjs, so from here only) and see if something needs to be done for supporting extensions?

Closing this as the method will have to be very different if so!

PR #174, opened 2023-05-23
1-to-1 Conversion of LZ-String to Typescript

karnthis:
Work in progress, creating draft for discussion and feedback during work. 

Completed:
- Core code
- base64-string
- Testing
- Bundling
- Config Refinement

Todo:
- Final Review


.gitignore
Rycochet, line 1: Not a blocker, but this sort of file should be in your home directory, as it's very specific to editors and dev environments - https://gist.github.com/subfuzion/db7f57fff2fb6998a16c :-)

README.md
```diff
@@ -1,5 +1,6 @@
-lz-string
+PLZSU
```
Rycochet, line 1:
```suggestion
lz-string
```

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "trailingComma": "none",
```
Rycochet, line 2: As TS is a code update, I'd want to switch this to `all` as part of the PR - it saves updating 2 lines to add another entry where one one is needed (won't comment on any files as this can be run automatically) :-)

```suggestion
  "trailingComma": "all",
```

src/interfaces/dictionary/Dictionary.ts
Rycochet, line 1: These are type only files, so can be named `*.d.ts` etc

src/main.ts
```diff
@@ -7,28 +7,31 @@
...
+const LZString: LZString = (function() {
 
 // private property
-var f = String.fromCharCode;
-var keyStrBase64 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=";
-var keyStrUriSafe = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+-$";
-var baseReverseDic = {};
+const f = String.fromCharCode;
+const keyStrBase64: string = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=";
+const keyStrUriSafe: string = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+-$";
```
Rycochet, line 18: Doesn't need the type hinting when it's being set explicitly :-)

```suggestion
const keyStrBase64 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=";
const keyStrUriSafe = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+-$";
```

src/main.ts
```diff
@@ -7,28 +7,31 @@
...
       baseReverseDic[alphabet][alphabet.charAt(i)] = i;
     }
   }
   return baseReverseDic[alphabet][character];
 }
 
-var LZString = {
-  compressToBase64 : function (input) {
+const LZString: LZString = {
+  compressToBase64 : function (input: string): string {
```
Rycochet, line 32: The type from the LZString should handle the arguments for these keys - but this also makes it very nicely specific! :-)

src/main.ts
```diff
@@ -7,28 +7,31 @@
...
     if (input == null) return "";
-    var res = LZString._compress(input, 6, function(a){return keyStrBase64.charAt(a);});
+    const res: string = LZString._compress(input, 6, function(a){return keyStrBase64.charAt(a);});
```
Rycochet, line 34: Known return type

```suggestion
    const res = LZString._compress(input, 6, function(a){return keyStrBase64.charAt(a);});
```

src/main.ts
```diff
@@ -38,97 +41,93 @@ var LZString = {
...
   //compress into uint8array (UCS-2 big endian format)
-  compressToUint8Array: function (uncompressed) {
-    var compressed = LZString.compress(uncompressed);
-    var buf=new Uint8Array(compressed.length*2); // 2 bytes per character
-
-    for (var i=0, TotalLen=compressed.length; i<TotalLen; i++) {
-      var current_value = compressed.charCodeAt(i);
+  compressToUint8Array: function (uncompressed: string): Uint8Array {
+    const compressed: string = LZString.compress(uncompressed);
+    const buf: Uint8Array = new Uint8Array(compressed.length*2); // 2 bytes per character
```
Rycochet, line 64:
```suggestion
    const compressed = LZString.compress(uncompressed);
    const buf = new Uint8Array(compressed.length*2); // 2 bytes per character
```

src/main.ts
```diff
@@ -38,97 +41,93 @@ var LZString = {
...
+    let context_dictionaryToCreate: PendingDictionary = {}
+    let context_c: string = ""
+    let context_wc: string = ""
+    let context_w: string = ""
+    let context_enlargeIn: number = 2 // Compensate for the first entry which should not count
+    let context_dictSize: number = 3
+    let context_numBits: number = 2
+    const context_data: string[] = []
+    let context_data_val: number = 0
+    let context_data_position: number = 0
```
Rycochet, line 122:
```suggestion
    let context_c = ""
    let context_wc = ""
    let context_w = ""
    let context_enlargeIn = 2 // Compensate for the first entry which should not count
    let context_dictSize = 3
    let context_numBits = 2
    const context_data: string[] = []
    let context_data_val = 0
    let context_data_position = 0
```

src/main.ts
```diff
@@ -323,30 +315,38 @@ var LZString = {
...
-        data = {val:getNextValue(0), position:resetValue, index:1};
+  _decompress: function (length, resetValue, getNextValue): string | null {
+    const dictionary: (number | string)[] = []
+    /** TODO - doesn't seem to be used. address during code cleanup */
+    // @ts-ignore
+    let next: number
+    let enlargeIn: number = 4
+    let dictSize: number = 4
+    let  numBits: number= 3
+    let entry: string = ""
```
Rycochet, line 332:
```suggestion
    let enlargeIn = 4
    let dictSize = 4
    let  numBits= 3
    let entry = ""
```

src/main.ts
```diff
@@ -397,14 +396,17 @@ var LZString = {
...
         return "";
     }
-    dictionary[3] = c;
+    /** TODO - address during code cleanup */
+    // @ts-ignore
+    dictionary[3] = c as string;
+    // @ts-ignore
     w = c;
+    // @ts-ignore
     result.push(c);
```
Rycochet, line 405: Not a 1:1 line but this can be cleaned significantly by making it a single line, what do you think? -

```suggestion
    result.push(dictionary[3] = w = String(c));
```

src/main.ts
```diff
@@ -460,14 +460,13 @@ var LZString = {
...
-
       if (enlargeIn == 0) {
         enlargeIn = Math.pow(2, numBits);
         numBits++;
       }
-
       if (dictionary[c]) {
-        entry = dictionary[c];
+        /** TODO - messy. address during code cleanup */
+        entry = dictionary[c] as string;
```
Rycochet, line 469:
```suggestion
        entry = String(dictionary[c]);
```

src/interfaces/dictionary/Dictionary.ts
```diff
@@ -0,0 +1,3 @@
+export interface Dictionary {
+  [key: string]: number
+}
```
Rycochet, line 3:
```suggestion
export type Dictionary = Record<string, number>;
```

src/interfaces/dictionary/DictionaryCollection.ts
```diff
@@ -0,0 +1,5 @@
+import { Dictionary } from '@/interfaces/dictionary'
+
+export interface DictionaryCollection {
+  [key: string]: Dictionary
+}
```
Rycochet, line 5:
```suggestion
export type DictionaryCollection Record<string, Dictionary>;
```

src/interfaces/dictionary/PendingDictionary.ts
```diff
@@ -0,0 +1,3 @@
+export interface PendingDictionary {
+  [key: string]: true
+}
```
Rycochet, line 3:
```suggestion
export type PendingDictionary Record<string, true>;
```

Rycochet (approved):
Need to change the single quotes for imports etc to use double quotes (including in config) - but looking good!!! :-D

.gitignore
karnthis: > Not a blocker, but this sort of file should be in your home directory, as it's very specific to editors and dev environments - https://gist.github.com/subfuzion/db7f57fff2fb6998a16c :-)

Sorry, I thought you were saying there was no .gitignore in the project root. On a reread I am confused now, are you saying the project should not have a .gitignore at all?

Rycochet: Generally I think it can be a mistake to having one that's nor relevant - but at the same time it removes a small amount of overhead for people wanting to contribute who don't have a global one :-P

.idea/.gitignore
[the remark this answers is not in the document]
Rycochet: That is to do with linting and formatting - so adding things like Prettier, eslint, and an .editorconfig - future PRs, and IDE independent :-)

.gitignore
karnthis: I don't see any harm in leaving it for those that don't have a global .gitignore. It doesn't conflict or prevent others from also having a global, while adding value to those that don't. If this sounds good, let's mark this item resolved.

src/interfaces/dictionary/Dictionary.ts
karnthis: afaik that is an anti-pattern unless you are documenting a js project by adding external types, and should only be used inline with the paired files (ie index.js and index.d.ts). Is that not correct?

karnthis:
> We know that should always be a string, but because at one point it might not be there's that ambiguity... I really don't like the thought of setting it as an EntityId type (string | number) - hence the explicit String(val) to force it to be the expected type... @pieroxy are you about to use your expertise on whether that could ever conceivably be anything other than a string? (search for the as string text for the few occasions)

My problem is not so much that it _could_ be different, but that is _is_ different, for example `const dictionary: (number | string)[] = []` in `_decompress` populates the first three index points with numbers, and the rest is entirely strings. Additionally main.ts line 423 `switch (c = bits)` is directly comparing what _should_ be a string to a number and later is set to a number on 439 for example.

Rycochet:
> My problem is not so much that it _could_ be different, but that is _is_ different, for example `const dictionary: (number | string)[] = []` in `_decompress` populates the first three index points with numbers, and the rest is entirely strings. Additionally main.ts line 423 `switch (c = bits)` is directly comparing what _should_ be a string to a number and later is set to a number on 439 for example.

The first one is easy then - `const dictionary: [number, number, number, ...string] = [];` (since they added rest into tuples), the second probably needs the cast (for now at least) ;-)

src/interfaces/dictionary/Dictionary.ts
Rycochet: Not sure on that - and no strong feelings - technically vite will make the right declaration files automatically, so it's purely a case of source code - could almost pull these all into a single file or leave them totally inline in a single source file (tbf I'm happy with anything external)

karnthis:
Ok, I completely forgot you can do that lol.

I have linting added, but I am leaving it out from the build process for now since I really want this move to be as explicit as possible for anyone that reads back later trying to understand the transition. Do we want to wait on @pieroxy to weigh in on the casting, or move forward on that? I'm not sure how active they are.

Rycochet:
I'd like to see @pieroxy give a thumbs up before merging - though maybe put a time limit of 2 weeks or so (basically I'm really busy until then - Friday June 9th to be arbitrary) - as soon as it's updated getting the linting and code-formatting sorted will be awesome!

karnthis:
> const dictionary: [number, number, number, ...string] = []

On further review, if we want to stick to 1:1 we will need to continue using:

```
const dictionary: (number | string)[] = []
```

for minimal change, or cast the first 3 to strings like so:

```
const dictionary: string[] = []
...
for (i = 0; i < 3; i += 1) {
      dictionary[i] = String(i);
    }
```

Using `const dictionary: [number, number, number, ...string[]] = []` would require redoing the existing code as you can''t declare specific elements on an empty array. It would yield something like this:

```
const dictionary: [number, number, number, ...string[]] = [0, 1, 2]
```

The above loop would be completely removed instead of modified.

package.json
```diff
@@ -1,42 +1,48 @@
...
+    "name": "pieroxy",
+    "email": "pieroxy@pieroxy.net"
+  },
+  "contributors": [
+    {
+      "name": "Karnthis",
+      "email": "me@erinrivas.com"
+    }
+  ],
   "version": "1.5.0",
```
Rycochet, line 13: Suggesting this as an alpha tag so people don't automatically update to it - once code formatting is sorted and we're happy with "this hasn't changed or broken anything including imports" then pull it to full release

```suggestion
  "version": "1.6.0-alpha.1",
```

Rycochet:
Could do it as `[] | [number] | [number, number] | [number, number, number] | [number, number, number, ...string]` though I'm not sure how TS would react to that? :-P

src/main.ts
karnthis: keeping until post-merge listing pass

src/main.ts
karnthis: keeping until post-merge listing pass

src/main.ts
karnthis: keeping until post-merge listing pass

src/main.ts
karnthis: keeping until post-merge listing pass

src/main.ts
karnthis: keeping until post-merge listing pass

package.json
```diff
@@ -1,42 +1,48 @@
...
-    "basePath": "libs/",
-    "files": [
-      "lz-string.js",
-      "lz-string.min.js",
-      "base64-string.js"
-    ]
+  "devDependencies": {
+    "@rollup/plugin-commonjs": "^25.0.0",
+    "@rollup/plugin-typescript": "^11.1.1",
+    "prettier": "^2.8.8",
```
Rycochet, line 40: One thing I don't like with prettier is how it ignores coding standards for "operators at the start of the line" - would you have any objections to the fork which actually does it correctly? (Pretty much the biggest issue on their repo) -

```suggestion
    "@btmills/prettier": "2.8.8",
    "prettier": "npm:@btmills/prettier@~2.8.8",
```

Related - remove the `^` prefix on versions (ie, lock them down for reproducibility), and add a `.npmrc` containing -

```
tag-version-prefix=""
save-prefix=""
```

src/interfaces/dictionary/Dictionary.ts
karnthis: updated per discussion

src/interfaces/dictionary/DictionaryCollection.ts
karnthis: updated per discussion

src/interfaces/dictionary/PendingDictionary.ts
karnthis: updated per discussion

package.json
karnthis: sure I'll flag it as an RC then.

package.json
karnthis: switched it out, I like that better. not going to run prettier until we are done because it touches nearly everything.

karnthis:
I did the casting as it seems like our cleanest option, but it will be super easy to remove if decided against.

.idea/.gitignore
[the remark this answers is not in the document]
karnthis: removed .idea from repo

karnthis:
Are we concerned at all with benchmarks at this time? Since it is just typing at this point I would be shocked if it yielded a meaningful difference.

Rycochet:
> Are we concerned at all with benchmarks at this time? Since it is just typing at this point I would be shocked if it yielded a meaningful difference.

Honestly I'll want to take a look at the final compiled output before allowing it to be merged - ideally we get a 1:1 library file, then can look at future improvements based on other PRs etc :-)

karnthis:
> Honestly I'll want to take a look at the final compiled output before allowing it to be merged - ideally we get a 1:1 library file, then can look at future improvements based on other PRs etc :-)
>

Do we want to commit the compiled code folder then? I was expecting we wouldn't and would set up a GitHub action to do the compile and npm publish. In my experience it is very uncommon to commit the dist folder.

Rycochet:
> > Honestly I'll want to take a look at the final compiled output before allowing it to be merged - ideally we get a 1:1 library file, then can look at future improvements based on other PRs etc :-)
> 
> Do we want to commit the compiled code folder then? I was expecting we wouldn't and would set up a GitHub action to do the compile and npm publish. In my experience it is very uncommon to commit the dist folder.

Given the locked package versions it should be pretty deterministic now - so I can compare locally when building :-)

karnthis:
Testing port is done, bundling is next.

did base64-string ever work? it has no tests and best I can tell doesn't actually function as-is.

karnthis:
bundling is done, tests added as sanity check for post-bundle code. I have done everything on my checklist, flagging as ready

Rycochet:
Just a ping to say I've not forgotten this - currently fighting against the uk heatwave (where AC isn't common, and I come from a far colder climate) - so hopefully get this checked out and merged later in the week! :-)

Rycochet:
So I've built it locally, run all the tests (passing, with around 95% coverage), and then compared the final output to the old final output (the old `lz-string.min.js` to the new `index.es.js`) where they are identical in terms of code (variable names change), but it doesn't have the UMD wrapper - which is included inside `index.umd.js`, with the correct package.json code to point bundlers at the right file.

So given that we've got a new RC release for it, and everything else being equal - enjoy!!!

Issue #182, opened 2023-11-28
Can utf-8 codes be avoided in the output?


marawannwh:
Hi, 

I am using this library in an ERP system called NetSuite. 

When I get an output from `LZString.compress` like `㚆ࡠ왐¸ὀ阁㌀룀☁᠄ᩈ恅唐聛ō瀿\x18ţ怂Ƅސȭ쀌먈\x8D퀁띌鴣Î\uDC71큣\x0E넕챞劜⑩逄켿搀ⴘɳ쏆` NetSuite saves it like `㚆ࡠ왐¸ὀ阁㌀룀☁᠄ᩈ恅唐聛ō瀿ţ怂Ƅސȭ쀌먈퀁띌鴣Î?큣넕챞劜⑩逄켿搀ⴘɳ`... and that makes the `decompress` function produce `null`.

Notice how the code `\x8D` for example was converted to `?`

Is there a way to avoid this other than using other methods that produce longer text?

Thanks

Rycochet:
Short answer: no

Longer answer: In order to compress to something it needs to be able to decompress it again. When there are "special" codes such as in utf8 then the library would need to understand them all for both sides of that. Unfortunately utf is an evolving standard, and that would mean having it work at the time of coding (probably!), and then break at some point in the future when new utf codes are added and they get special meaning (or conversely unknown codes that get swapped with `?` get a glyph and suddenly start behaving differently).

If your code is getting z-lib compressed (ie, browser compression) then it won't make much difference using base64.

marwan-nwh:
I tried `compressToUTF16`, and in the cases I ran it through it didn't produce codes like that. So this doesn't mean it will never produce UTF8 codes, right? 

What if I replace `\x` before saving with something unique, then change it back before decompressing? I think this would work if there are some characters that are guaranteed NOT to be generated in the compression output by the lib. Are there such characters?

HelloLudger:
compressToCustom and decompressFromCustom should work for your usecase, if you can provide the string of allowed characters.

marwan-nwh:
> compressToCustom and decompressFromCustom should work for your usecase, if you can provide the string of allowed characters.

I tried it, but the length of the output is larger than `compressToEncodedURIComponent`'s output. I thought it would be as small as `compress` while changing utf-8 codes with one of the characters I am passing.

Rycochet:
[...]

Issue #180, opened 2023-10-16
Feature request: compress to source-compatible character set


emlyn:
It would be nice to have a mode that can be dumped as a string in a javascript source file (I think UTF16 without special characters below 0x20 and without single or double quotes... I'm not sure if there are other characters that could be problematic).

Issue #169, opened 2023-02-24
Is there a way to allow custom dictionaries?


wll8:
For example, add a parameter: dict, when the value is ```ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789!#$%&()*+,./:;<=>?@[]^_`{|}~"```, it means use base91 compression. 

see: #90 

It looks like base91, but it is not the standard base91 algorithm, but only uses the characters of these sets as output.

When the value is `hello lz`, it means that only some combinations of these characters will be included in the compressed text.

pieroxy:
Hello. I am not sure what your question is or what you are trying to achieve. Could you maybe explain a bit more ?

wll8:
It seems that this is a function of converting a string to only contain certain characters specified.

Example: Pseudocode


```js
lz.compress(`hello`, {
   map: `0123456789ABCDEF`,
});
```
The output of the above code is: `68656C6C6F`, they do not contain characters outside the map.

Can lz-string provide a way to alleviate the size increase by using the lz algorithm during the conversion process?

Here are some more examples:

pseudocode: looks like morse code

```js
lz.compress(`hello`, {
  map: `. -`,
});

// .... . .-.. .-.. ---
```

pseudocode: looks like Base16

```js
lz.compress(`hello`, {
  map: `0123456789ABCDEF`,
});

// 68656C6C6F
```

pseudocode: Array

```js
lz.compress(`hello`, {
  map: ["公正", "爱国", "平等", "诚信", "文明"],
});

// 公正爱国公正平等公正诚信文明公正诚信文明公正诚信平等
```

pseudocode: Array

```js
lz.compress(`hello`, {
  map: `👨👩👧`,
});

// 👨👨👧👨👨👧👨👧👩👨👧👨
```

pieroxy:
> Can lz-string provide a way to alleviate the size increase by using the lz algorithm during the conversion process?

This is very exactly what LZ-String is doing. It is using the lzw algorithm to compress the input, using the symbols in the array to store the bits the algorithm produced. So in effect, `lz.compress('hello', {map: '0123456789ABCDEF'});` does not look like base16, it *is* a base16 representation of the compressed stream.

wll8:
You have provided the following functions, but none of them meet the requirements, because in my use case only specified characters are allowed to appear in the compression result.

- compress
- compressToUTF16
- compressToBase64
- compressToEncodedURIComponent
[...]

PR #183, opened 2023-12-10
Added compressToCustom and decompressFromCustom function

HelloLudger:
as discussed in / (maybe) useful for
- #169
- #90 
- #165 
- #180 
- #182 

It basically works similar to `...Uint8Array()`, first compressing everything and then figuring out, how the UTF-16 characters can be stored with the provided dictionary(-string).

I couldn't make the `_compress`/`_decompress` route work, because the custom dictionary length is often not 2^something, so `_decompress` wouldn't work (or `_compress` would only use a subset of the provided characters).


Rycochet:
Can this be updated to the new split file version - should be relatively easy to do - testing is getting improvements but the basic pattern is in there!

HelloLudger:
Yes, looking into it. Hopefully this weekend.

HelloLudger:
Hey, I hope this will work for you.

tests/testValues.ts
```diff
@@ -112,3 +112,17 @@ export const test_tattooUint8Array = new Uint8Array([
...
+
+/** A longer text including very unusual symbols */
+export const test_complicated = `Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.
```
Rycochet, line 117: I'm thinking through updating the tests, and the data used - this is a good idea, however it doesn't need the symbols - unicode is tested elsewhere so having a reproducible source text is good though :-)

tests/testValues.ts
```diff
@@ -112,3 +112,17 @@ export const test_tattooUint8Array = new Uint8Array([
...
+    Ut wisi enim ad minim veniam, quis nostrud exerci tation ullamcorper suscipit lobortis nisl ut aliquip ex ea commodo consequat. Duis autem vel eum iriure dolor in hendrerit in vulputate velit esse molestie consequat, vel illum dolore eu feugiat nulla facilisis at vero eros et accumsan et iusto odio dignissim qui blandit praesent luptatum zzril delenit augue duis dolore te feugait nulla facilisi.   
+    Nam liber tempor cum soluta nobis eleifend option congue nihil imperdiet doming id quod mazim placerat facer possim assum. Lorem ipsum dolor sit amet, consectetuer adipiscing elit, sed diam nonummy nibh euismod tincidunt ut laoreet dolore magna aliquam erat volutpat. Ut wisi enim ad minim veniam, quis nostrud exerci tation ullamcorper suscipit lobortis nisl ut aliquip ex ea commodo
+    Alle Unicode-Symbole und ihre Codes auf einer Seite, 𐎰, 103B0, 𐎱, 103B1, 𐎲, 103B2, 𐎳, 103B3, 𐎴, 103B4, 𐎵, 103B5, 𐎶, 103B6, 𐎷, 103B7, 𐎸, 103B8, 𐎹, 103B9, 𐎺, 103BA, 𐎻, 103BB, 𐎼, 103BC, 𐎽, 103BD, 𐎾, 103BE, 𐎿, 103BF, 𐏀, 103C0, 𐏁, 103C1, 𐏂, 103C2, 𐏃, 103C3, 103C4, 103C5, 103C6, 103C7, 𐏈, 103C8, 𐏉, 103C9, 𐏊, 103CA, 𐏋, 103CB, 𐏌, 103CC, 𐏍, 103CD, 𐏎, 103CE, 𐏏`;
```
Rycochet, line 128: For safety this unicode should always be escaped - ie, don't trust the editor so the first should be `\u103B0` etc

src/custom/custom.test.ts
```diff
@@ -0,0 +1,59 @@
...
+import { compressToCustom, decompressFromCustom } from ".";
+import { test_longString_fn } from "../__tests__/testFunctions";
+import {
+    test_hw,
+    test_repeat,
+    test_tattooSource,
+    test_complicated,
+} from "../../tests/testValues";
+
+describe("custom", () => {
```
Rycochet, line 11: Can you follow the pattern of other tests - make an arrow function for each of the source dictionaries, the long text can be added to the main testFunctions so all the compression functions can be tested against it :-)

Rycochet:
Only thing blocking this is getting the tests to use the same boilerplate as elsewhere - trying to figure out how to add this to the commandline code (still working on that so don't worry about adding it yourself!)

src/custom/custom.test.ts
HelloLudger: Sorry, I thought I followed the pattern I saw in `encodedURI.test.ts`.
I am not sure what I should change here.

HelloLudger:
I didn't use `runTestSet` because I don't know how to deal with the second parameter for the custom function (the custom dictionary). The "lazy" way would be to give them a default custom dictionary, then `runTestSet` could be used as is.

tests/testValues.ts
HelloLudger: Wouldn't change that the test case? 
Or will every system deal with `\u103B0` the same way it deals with 𐎰?

src/custom/custom.test.ts
Rycochet: Ah, I get what you mean - the proper way to do it would be something like this -

```ts
describe("custom", () => {
	describe("hex", () => {
        const hex    = "0123456789ABCDEF";
        const compress = (input) => compressToCustom(input, hex);
        const decompress = (input) => decompressFromCustom(input, hex);

        runTestSet<string>(compress, decompress, "");
    });

	describe("base62", () => {
        const base62 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
        const compress = (input) => compressToCustom(input, base62);
        const decompress = (input) => decompressFromCustom(input, base62);

        runTestSet<string>(compress, decompress, "");
    });
});
```

The nested `describe` groups by type, and the arrow function lets you pass that second argument transparently - hope that makes more sense :-)

tests/testValues.ts
Rycochet: Javascript itself decodes it into real unicode internally when reading the string - but while we're probably all using editors that handle unicode internally properly there's no guarantee of that (or even that there isn't some global `.git` option for someone else that stops apparent text files from having the upper bit cleared as it's ascii etc) - being defensive is better (and also reads far easier in case someone doesn't have a unicode compatible font)

Rycochet:
> I didn't use `runTestSet` because I don't know how to deal with the second parameter for the custom function (the custom dictionary). The "lazy" way would be to give them a default custom dictionary, then `runTestSet` could be used as is.

I've replied on the thread, hopefully it makes more sense now :-)

Rycochet:
Also note that we've got a consistent license header on all other files now, can you add that to your new ones :-)

HelloLudger:
Next try. Hopefully it works now. Found a missing typing for compressToCustom. Seems like TypeScript and testing is useful after all.

I tried to create a test case with an 13-character heterogram (`unproblematic`), but then the tests "Repeating String" and "Long String" fail, because the "compressed" string is longer than the original string. The custom dictionary has to have at least 16 chars to pass these tests.

Just FYI, for the testing it is not important.

src/custom/custom.test.ts
```diff
@@ -0,0 +1,35 @@
...
+        runTestSet<string>(compress, decompress, "");
+    });
+
+    describe("23 letter heterogram", () => {
+        const heterogram = "BlockyDwarfZingsTheJump";
+        const compress = (input: string | null) => compressToCustom(input, heterogram);
+        const decompress = (input: string | null) => decompressFromCustom(input, heterogram);
+
+        runTestSet<string>(compress, decompress, "");
+    });
```
Rycochet, line 26: This is awesome!!

tests/testValues.ts
Rycochet: I've got a new chunk of Lorem Ipsum that will be coming in later today - but this is good until then!

Rycochet (approved):

Rycochet:
The size test is probably not important - I'm moving over to only testing known output and making sure that the raw data is hashed and known correct - should have that merged later tonight, which will give us a better testing base :-)

marwan-nwh:
Sorry, I am a little bit confused. How to use this modification? I thought it added `compressToCustom` and `decompressFromCustom` to the lib, but I just installed it, and these functions don't exist.

Rycochet:
@marwan-nwh How did you install it? v2 (that this is part of) is not currently released to npm so you need to install manually.

marwan-nwh:
> @marwan-nwh How did you install it? v2 (that this is part of) is not currently released to npm so you need to install manually.

Got it, thanks!

HelloLudger:
Can you elaborate on the found problem? Anything I can do to help?

Rycochet:
@HelloLudger I've created a draft PR with more detail and an easier way to test it - https://github.com/pieroxy/lz-string/pull/200 - hope that helps!

Issue #184, opened 2023-12-15
`decompressFromBase64` can return `null`


emewjin:
version: 1.5.0

[...]

Issue #177, opened 2023-06-20
Version 2 TODO list


Rycochet:
Non-exhaustive list of things to do before 2.0.0 can come out of RC status

- [x] Re-enable command line binary
- [x] Cleanup code for more modular approach
- [x] Extend and update tests to make them easier to understand
- [x] Ensure we have full AMD support
- [ ] Enable Github Actions for release, eslint, etc
- [ ] Add "corrected" base64 endpoint (separately named endpoint for #110 )
- [ ] Add "corrected" encodedURI endpoint (separately names endpoint for #127 )

... Other issues and bugfixes can be added, however the important thing for this initial v2 release is maintaining code compatibility, more major improvements should wait for future versions (think of the v2 releases as the in-between "we might deprecate some endpoints, but they'll still exist")

karnthis:
a thought, GitHub actions should be added to automatically publish releases to NPM. currently the RC.0 is documented as available via npm install, which it is not.

I feel we should identify a list of which 3rd party implementations we want to test against for compatibility. other versions are listed in the readme, but are noted as may not be 100% compatible. I also have not reviewed how out of date projects may or may not be, and as 2.0 is an attempt to modernize the codebase significantly I can pretty much promise that something is going to permanently break. perhaps the solution is a legacy mode of some sort? it feels like it could be a significant duplication of code though.

final thought, I would really like to see this move to an instantiated class model. I feel it would significantly open our options, but would be a major breaking change as everyone would have to init the class now instead of directly using the import.

karnthis:
I was doing some poking around, and noticed that the original lzstring file had `define.amd` in it. I just used defaults when setting up site, so AMD in not included. this is an easy-ish change, do we want to do that or is that legacy enough for us to drop it?

Rycochet:
I'm wanting to set up github actions for it - had a quick look when I had time (but going to be another busy week for me!) - it did look pretty simple, but there needs to be some other things like branch protection, and requirements for eslint etc to pass before merging - all nice actions to add!

Before 2.0 final is released we definitely need to have full compatibility in there so anyone upgrading should have an pretty transparent upgrade path (with some nice `@deprecated` notes in the Typescript and docs appearing over time).

[...]

Issue #110, opened 2017-12-14
Somehow compressToBase64() returns string end with 3 equal sign "==="


linkenmin:
Somehow compressToBase64() returns string end with 3 equal sign "===", which is not a valid format by base64 definition.
Does this behavior meet your expectation?
Thank you.

Example source:
abcdef

Result:
IYIwxgJgpgZkA===

nickion:
The faulty code is
`switch (res.length % 4) { // To produce valid Base64 `
and the following switch statement. It should be a mod 3 with cases for 0, 1 and 2.

pieroxy:
Nice catch, thanks. Will fix soon.

nickion:
and thanks for the library; spent a while looking and testing libs for compressing local storage, and yours was by far the best overall that I found.

GeekStocks:
Yes @pieroxy we love this lib too — will likely implement it when this Base64 bug is squashed. Use the Base64 in testing alot, thanks @linkenmin for the catch and @nickion for the fix!

AmericanY:
```python
import lzstring


x = "N4IgLgngDgpiBcIBGAnAhgOwCYgDQgBs0EQYM8QBHASxIAYBaGAOSwAUAOAMzAHY0AYgHcAWtQCuADQD2WNAQBeADyRIhCgIIBFLABlpANQIARAEIBNABIQAVlwCi0gKoBZALwVK4mN4QBGfAB9Ej8/Og46EABfIAAA="

z = lzstring.LZString().decompressFromBase64(x)

y = lzstring.LZString().compressToBase64(z)

print(y)

print(x == y)
```

The padding is incorrect!

AmericanY:
@nickion Hello, could you please help me out to solve that?

pieroxy:
LZString works the other way: For any string, if you compress it and then decompress it, the original and resulting strings will be equal.

The padding changed in a version (I don't remember which one to be honest). What are you trying to achieve here ?

AmericanY:
@pieroxy Thanks for your response. Actually, am trying to decompress and compress the value to be equal together. as am trying to call an [API](https://stackoverflow.com/questions/66697604/hard-scraping-api) with those value which require lzstring compress

AmericanY:
@pieroxy solved by installing `pip install "lzstring==1.0.3"`

pieroxy:
Great. There's probably a tiny bit of loss in performance, but this should be nothing to worry about.

[...]

PR #189, opened 2024-01-17
Issues and v2 work updates

karnthis:
Work to reduce outstanding issues new and old. Impacted items include:
- #184
- #177
- #110
- #127

List to be updated as work is completed.


src/utils/misc.ts
```diff
@@ -0,0 +1,14 @@
...
+    thing: string,
+    version: VERSIONS,
+    opts?: { aggressive?: boolean; replacement?: string }
+): void {
+    let notice = `LZString | ${thing} is deprecated as of: ${version}`
+    if (opts?.replacement) {
+        notice += ` - Please use ${opts.replacement} instead`
+    }
+    console.error(notice)
+    if (window && opts?.aggressive) alert(notice)
```
Rycochet, line 13: This isn't built for the browser, so not sure exactly what to do here, but maybe adding a `declare var window` or similar - not sure that this is the best thing to do blocking everything as some people might update without checking etc

karnthis: Fair enough, it is simple to take it out. Wanted to cover all the bases and with all the benchmarking pages that have been done I thought browser support was needed.

Rycochet: It's needed - but TS / vite doesn't know it's there - so need to make sure that this doesn't throw an error when it doesn't exist :-)

karnthis: If you like, we can add "DOM" back to the lib compilerOptions in tsconfig.json so ts/vite stops throwing compile errors.

Rycochet: That gives too much risk of something coming in that shouldn't - if the only place that exists is in this file then it feels safe - technically could disable the check on this line, but better to do it "properly" when it's a known :-)

karnthis: Removed browser-specific elements. We can always add the flags back later when we have a proper spec for browser support.

package.json
```diff
@@ -52,7 +52,8 @@
         "build": "tsc && vite build",
         "format": "prettier --write .",
         "start": "vite build --watch",
-        "test": "[ -f ./dist/index.js ] || npm run build && vitest",
+        "test": "[ -f ./dist/index.js ] || npm run test:build",
+        "test:build": "npm run build && vitest",
```
Rycochet, line 56: This isn't doing quite what you expect - previously it was running the build if it hadn't run already - we actually don't need that conditional run, so can just have `vitest` on its own - I was mixing up `vite preview` and `vitest`!

```suggestion
        "start:test": "vitest",
        "test": "vitest run",
```

karnthis: `run build` is required to capture changes to the codebase because tests run specifically against the files in /dist, and without it the test system will only reflect changes to the tests themselves. `test:build` forces a build of the codebase to ensure all tests run successfully against the latest code.

if there is another way to capture code changes in list I would love to use it because this honestly isn't great.

Rycochet: Good point - it still needs that ternary for the moment then - I'm about to be pushing a test update that tests the built cjs (currently) via the new commandline tool and a bash script (the intention is that other implementations can support the same arguments and then run the same tests) - currently I can't get nodejs to load binary data properly though, so raw and uint8array both fail to decode :-(

karnthis:
@Rycochet your thoughts, if you would please.

package.json
Rycochet: All the test code has been updated and fixed, so it works properly - the test strategy will be changing slightly for the production code so it'll only run when it can (and force it on github itself) - so I'll clean up this when I get to it - can undo the change in the meantime :-)

src/__tests__/testFunctions.ts
Rycochet, line 1: Not sure what the conflicts are but update the file and I'll take a look :-)

src/__tests__/testFunctions.ts
```diff
@@ -39,6 +39,39 @@ export function getTestData(name: string) {
     return cachedTestData[name] || (cachedTestData[name] = readFileSync(`testdata/${name}/data.bin`).toString());
 }
 
+// eslint-disable-next-line  @typescript-eslint/no-explicit-any
+function runGeneralTests(identifier: string, compressFunc: (input: any) => any, decompressFunc: (input: any) => any) {
```
Rycochet, line 43: Typedocs to be added, and use generics rather than `any` - but abstraction is good :-)

src/utils/misc.ts
```diff
@@ -0,0 +1,9 @@
+type VERSIONS = "v2.0.0";
+
+export function deprecated(thing: string, version: VERSIONS, opts?: { replacement?: string }): void {
```
Rycochet, line 3: Cache whether the notice has been written for this function before, otherwise it can spam the console.

I'd actually suggest making this into a decorator instead which makes the code "cleaner" - it also means you can replace the function call with the original after the first call so no need to cache it (hmu if you need to!)

I've opened a poll on #190 which will decide whether to do this style, or fix them and provide legacy endpoints (I'm leaning towards this personally, and it's the more "lz-string" way, but  if we get a strong opinion another way I'll happily go with it)

src/base64/compressToBase64URL.ts
Rycochet, line 1: I think this should be a separate folder - `base64URL/` (same for decompress) - it makes for a cleaner direct import

src/base64/compressToBetterBase64.ts
Rycochet, line 1: `compressToBase64_fixed` (for now)

src/base64/decompressFromBetterBase64.ts
Rycochet, line 1: `compressFromBase64_fixed` (for now)

src/__tests__/testFunctions.ts
```diff
@@ -84,36 +117,46 @@ export function runTestSet<T extends { length: number }>(
...
-                    // @ts-expect-error We don't know the type
-                    expect(decompressFunc(knownCompressed)).toEqual(rawData);
-                });
-            }
-        });
-    }
+        expect(compressedEmpty).toEqual(compressFunc(""));
+        expect(compressedEmpty).toEqual("");
+        compressedEmpty instanceof Uint8Array
+            ? expect(compressedEmpty.length).not.toBe(0)
```
Rycochet, line 154: The length property isn't necessarily as intended (and if we extend to other non-string types it's not necessarily safe), so TypedArray `byteLength` instead?

```suggestion
            ? expect(compressedEmpty.byteLength).not.toBe(0)
```

src/__tests__/testFunctions.ts
```diff
@@ -84,36 +117,46 @@ export function runTestSet<T extends { length: number }>(
...
+            : expect(typeof compressedEmpty).toBe("string");
```
Rycochet, line 155: It's got a choice of two types, you're checking one to check the size, and then the other to check the type - better to separate out the length check from the type check (and check that it's a supported type so `Uint8Array` is also valid) :-)

Rycochet:
Almost there I think :-D

src/__tests__/testFunctions.ts
karnthis: I basically just copied over from the original and didn't want to change too much, but I just cleaned it up a bunch.

package.json
Rycochet: This can now be removed - the test stategy has been updated (including better tests for built code, and an external test tool [here](https://github.com/pieroxy/lz-string/tree/master/test) - I'll get it running in github actions soon!)

src/__tests__/testFunctions.ts
Rycochet: I'm sure it's possible, but might be worth merging this in and fixing later

src/base64/base64.test.ts
```diff
@@ -5,9 +5,13 @@
  */
 
 import { describe } from "vitest";
-import { compressToBase64, decompressFromBase64 } from ".";
-import { runTestSet } from "../__tests__/testFunctions";
+import { compressToBase64, compressToBase64_fixed } from ".";
+import { decompressFromBase64, decompressFromBase64_fixed } from ".";
```
Rycochet, line 9: Should merge into one import (format might want to play though!)

```suggestion
import {
    compressToBase64,
    compressToBase64_fixed,
    decompressFromBase64,
    decompressFromBase64_fixed,
} from ".";
```

src/utils/misc.ts
Rycochet: I started to write a reply and go through the various ways it could be done (as we can't use a real typescript / javascript decorator outside of a class) - and realised that it's premature optimisation and we can worry about this later - plus it's far more useful for developers than for the end-user, so we've got until the final release of v2 to improve matters, with this as a good place to start from :-)

Rycochet (approved):
I think this is good to merge once updated and tests are run :-)

karnthis:
I have everything done except the tests. since you redid the library I am having issues as my tests no longer work but the code being tested did not change. @Rycochet can you go into details on your process for setting up the base64 tests, since that is very similar to what the new tests need?

Rycochet:
@karnthis The biggest thing is that the tests *will fail* when there is a fix to the underlying code - they're comparing against a "known good" binary output, but in reality the fix means that's not correct any more - you should be able to delete the files for the two fixed encoders, and re-run the `test.sh` script with the `-u` option to update those files - which will then be used by the actual `vitest` tests - do feel free to update things and push with those tests still failing if it's not clear (and then I can checkout locally and give any other pointers that might be needed) :-)

Rycochet:
(Note that once this is merged I want to add specific handling for the legacy versions into the test scripts - but not entirely sure what form that needs to take!)

ghost (changes requested):
Deny

karnthis:
considering the significant number of changes that have happened to the master branch since this started I am considering scrapping what is here and starting fresh. The work still needs to be done but it will need to be approached differently now that things have stabilized a bit.

@Beeloo0011 what?

wkrick:
> Beeloo0011 what?

@karnthis 

I think this is a bot.  Their GitHub history is really bizarre.  Like they're probing for insecure repos or something.

karnthis:
yeah that was my assessment too, was curious to see if I would get a response. didn't expect someone else to reply lol.

ghost (approved):

karnthis:
@Rycochet

PR #215, opened 2024-02-21
Replace objects with Maps to ensure compatibility with Hermes

danii1:
This PR replaces objects with Map to alleviate the issue with Hermes runtime(React Native) that doesn't support large number of props in object.

### DIFF
--- a/libs/lz-string.js
+++ b/libs/lz-string.js
@@ -13,16 +13,16 @@ var LZString = (function() {
 var f = String.fromCharCode;
 var keyStrBase64 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=";
 var keyStrUriSafe = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+-$";
-var baseReverseDic = {};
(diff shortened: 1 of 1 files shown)


Rycochet:
This needs to compare with #98 and have a look at any side effects etc mentioned - not got time to review myself now, but of the general opinion that it's worth adding :-)

danii1:
I have another branch with the same changes applied to v2: https://github.com/DISCO-team/lz-string/tree/feature/object-map

Ran test:bench, Map seems faster in most scenarios(particularly where it matters like compressing large text), not surprising given it's optimized for addition and deletion:
[obj.txt](https://github.com/pieroxy/lz-string/files/14363398/obj.txt)
[map.txt](https://github.com/pieroxy/lz-string/files/14363397/map.txt)

adbs1:
@Rycochet @danii1 Any further progress likely here? The changes look trivial with a cursory glance :)

karnthis:
The change is solid and has my vote of support from a code standpoint, but needs a final decision from a compatibility angle. LZ-String has been maintaining compatibility with anything javascript released since roughly 2012, and this change will break that trend. I believe there is a limit to backwards compatibility and a move to v2 is a good time to revisit that support, but is ultimately not my decision to make. @Rycochet @pieroxy what are your thoughts?

PR #269, opened 2024-09-02
Make lz-string tree-shakable for bundlers

y-hiraoka:
**Description:**

This pull request introduces a small but impactful change to the `lz-string` library to make it tree-shakable by modern bundlers. By modifying the entry point structure, this change enables a significant reduction in bundle size—over 30% when using esbuild without minification.

**Details:**

- **Before:** `import defaultExport from "lz-string"`
- **After:** `import * as defaultExport from "lz-string"`

This adjustment allows bundlers to effectively eliminate unused code, leading to leaner bundles. When using a single `compress`/`decompress` pair, this resulted in a 30% reduction in size. Although this change slightly alters how the library is imported, it offers considerable benefits in terms of performance and efficiency.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -6,7 +6,7 @@
 
 import { describe, test } from "vitest";
 
-import index from "../index";
+import * as index from "../index";
 
 describe("index.ts", () => {
     test("was the change deliberate?", ({ expect }) => {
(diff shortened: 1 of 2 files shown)


y-hiraoka:
The repo appears to be unmaintained, so I'm closing this pr.

